PR #56, opened 2014-04-27
Multiple apps for ZAT.

ocke:
In order for this to work you need the ZAS that supports multiple apps too.
The commit will add support for loading multiple apps in zat.

zat server app_path1 app_path2 app_path3

This code is backwards compatible with existing code.


digitalmaster:
Was this ever merged?? Would be so useful to test two apps at the same time +1

ocke:
@digitalmaster we moved it: https://github.com/zendesk/zendesk_apps_tools/pull/105

PR #75, opened 2015-01-04
API endpoint for deploy#find_app_id seems to have been incorrect or old ...

jpalmieri:
...(and was causing a zat update to fail), so I updated it to '/apps/owned.json' per the documentation


adammw:
Strange, as far as I know, `/api/v2/apps.json` **should work** however it will list both public apps and the private apps you've uploaded, compared to `/api/v2/apps/owned.json`.

/cc @zendesk/quokka

princemaple:
Please reopen if the problem is still there. We believe that it's not a problem. I personally use `zat update` fairly often.

jpalmieri:
Sorry, meant to update this. The `/api/v2/apps.json` endpoint is working for me again; it must have been down for some reason when I submitted this.

princemaple:
Good to know, thanks!

PR #76, opened 2015-01-14
Watch the config file (yml/json)

ocke:
The sinatra server now checks the config file time on every load of app.js.
If the file has been updated, it will reread the config file.

I made this because messing around with the config file, I constantly had to restart the zat server (ctrl-c, arrow up, enter). And because I'm very lazy, I don't want to do that every time.


lib/zendesk_apps_tools/server.rb
```diff
@@ -4,9 +4,20 @@
...
 
     get '/app.js' do
       content_type 'text/javascript'
+
+      if (File.exists? settings.config)
+        curr_mtime = File.stat(settings.config).mtime
+        if (curr_mtime > last_mtime)
+          settings_helper = ZendeskAppsTools::Settings.new
+          settings.parameters = settings_helper.get_settings_from_file settings.config, settings.manifest
+          last_mtime = File.stat(settings.config).mtime
```
adammw: any reason you're not setting it to the cached mtime (`curr_mtime`) but looking it up again?

lib/zendesk_apps_tools/server.rb
```diff
@@ -4,9 +4,20 @@
 module ZendeskAppsTools
   class Server < Sinatra::Base
     set :public_folder, Proc.new {"#{settings.root}/assets"}
+    last_mtime = Time.new(0)
 
     get '/app.js' do
       content_type 'text/javascript'
+
+      if (File.exists? settings.config)
```
princemaple: unnecessary brackets

lib/zendesk_apps_tools/server.rb
```diff
@@ -4,9 +4,20 @@
...
+        curr_mtime = File.stat(settings.config).mtime
+        if (curr_mtime > last_mtime)
```
princemaple: unnecessary brackets

lib/zendesk_apps_tools/server.rb
```diff
@@ -4,9 +4,20 @@
...
+          settings_helper = ZendeskAppsTools::Settings.new
+          settings.parameters = settings_helper.get_settings_from_file settings.config, settings.manifest
```
princemaple: actually prefer brackets when doing method call

princemaple:
:+1:

adammw:
:+1:
I found this code rather hard to read though as I saw about three different meaning for the variable called 'settings'.

lib/zendesk_apps_tools/server.rb
ocke: oops. missed that one. Will fix

ocke:
haha, I thought the same thing. The server 'settings' are called settings and those have app settings...

If you have a better naming idea, be my guest to change (or suggest) it. :+1:

ocke:
@svizzari Any thoughts guys?

svizzari:
:+1: I'm fine for this PR to go ahead and be merged when you can resolve merge conflicts with master

ocke:
@svizzari done.

sandlerr:
Thank you!

PR #78, opened 2015-02-02
[LAB][POC] Testable apps poc

danielbreves:
:koala: :construction: 

/cc @zendesk/quokka

zat_test

### To do:
- [ ] Remove method calls from installed.js 

### References
 - Jira link: https://trello.com/c/lM7tcicY/11-make-apps-testable-possibly-using-zat

### Risks
 - None

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,6 +18,7 @@ PATH
   specs:
     zendesk_apps_tools (1.21.1)
       faraday (~> 0.8.7)
+      jasmine (~> 2.1.0)
       rubyzip (~> 0.9.1)
       sinatra (~> 1.3.4)
       thor (~> 0.18.0)
@@ -53,6 +54,12 @@ GEM
       multi_json (~> 1.3)
     i18n (0.6.11)
     image_size (1.3.1)
+    jasmine (2.1.0)
+      jasmine-core (~> 2.1.0)
+      phantomjs
+      rack (>= 1.2.1)
+      rake
+    jasmine-core (2.1.3)
     jshintrb (0.2.4)
       execjs
       multi_json (>= 1.3)
@@ -62,7 +69,8 @@ GEM
     multi_json (1.10.1)
     multi_test (0.1.1)
     multipart-post (1.2.0)
-    rack (1.5.2)
+    phantomjs (1.9.8.0)
+    rack (1.6.0)
     rack-protection (1.5.3)
       rack
     rake (10.3.2)
--- a/jasmine.yml
+++ b/jasmine.yml
@@ -1,0 +1,123 @@
+# src_files
+#
+# Return an array of filepaths relative to src_dir to include before jasmine specs.
+# Default: []
+#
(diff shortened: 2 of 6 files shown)


adammw:
So we only support jasmine tests for this POC?

atroche:
so awesome!

atroche:
@adammw it'd still be a massive improvement over having no testing capabilities

danielbreves:
@adammw yes, jasmine was the simplest to setup.

ocke:
Any change for mocha?

svizzari:
> chance

? haha, I knew you'd ask that @ocke.  We've been discussing around test framework options but wanted to proof the concept first as quickly as possible.  What do you see as the advantages of Mocha over Jasmine?

ocke:
All my field-marshal2, quickie and bubbly tests are in mocha. So the advantage is that I don't have to rewrite anything and can spent time make zendesk_apps_tools more awesome! ;-)

damienwhaley:
Please allow me chip in here for my 2c. Jasmine and Mocha are very similar. Mocha is slightly newer than Jasmine and suits Node.js a bit better (especially async tests). The main difference is the expect or assert tests. Mocha works with should or chai, whereas Jasmine has expect built in.

We've been using Mocha as most of the testing we've been writing has been for Node.js, and the testing we have been doing with he require modules can be run with Node.js and Mocha.

Both Mocha and Jasmine can also be run as tests in the browser which is what has been implemented in this pull request.

The winner for ZAT at the moment is Jasmine as there's a Ruby Gem for it - and there's already a Proof of Concept in which it is working.

So I think we would be OK with Jasmine as the testing framework, and it is not too painful to move between the two frameworks.

The ultimate solution would be to be able to run tests while the app is running in Zendesk. We've written a couple of proof of concept apps that do this in different ways, but have not come up with a good solution yet. 

The Proof of Concept is a good first step and for what it's worth I think it's a good thing.

It would be good if the files in the spec directory were included in the app which is uploaded to Zendesk, and that they would be available to be required (I would think that this would enable us to have a way of "self-testing" apps running in Zendesk).

svizzari:
@damienwhaley Thanks for the response :smile_cat: 

We've used both Jasmine and Mocha in the past across Zendesk projects.  I was curious as to whether there was a particular feature that was a deal-breaker for @ocke / L&L.

The pluggable nature of Mocha and the ability for the developer to choose the assertion or mocking library that suits them is nice.  On the other hand, Jasmine makes it easy to get up and running as the DSL includes a full suite of features.  IMO, Jasmine 2.0+ has a more Mocha like DSL and has much improved (over 1.3) for async testing.

We've been having some discussions around how it might be possible to run browser tests for an app from ZAT, but haven't landed on a solution as yet. We'll definitely keep you in the loop on that possibility.

danielbreves:
Closing this until we can reduce the dependencies on installed.js.

svizzari:
:crying_cat_face:

PR #79, opened 2015-02-09
Allow setting of the app id

adammw:
:koala:

Allow setting of the app id so we can use ZAT for testing existing installations of apps that use Secure Requests without having to re-upload them. ZAS uses the same id for the installation as well, so oddly you'll usually want to use the installation id and not the app id for the app id. :/

/cc @zendesk/quokka

### Risks
 - None

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -88,11 +88,13 @@ module ZendeskAppsTools
     DEFAULT_SERVER_PATH = './'
     DEFAULT_CONFIG_PATH = './settings.yml'
     DEFAULT_SERVER_PORT = 4567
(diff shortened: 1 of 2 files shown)


princemaple:
:+1:

atroche:
cool :+1:

svizzari:
Nice @adammw

danielbreves:
@adammw cool! Though shouldn't we call this installation_id instead?

danielbreves:
@adammw what happens with settings? Could it skip prompting the user for settings and use the installation settings instead?

adammw:
@danielbreves that's not a bad idea, but for "app id" it doesn't necessarily make as much sense. Sounds like a natural extension though of the idea.

PR #81, opened 2015-02-18
Add the ability to bump app version

princemaple:
:koala: 

Usage:

```
zat bump major
zat bump minor
zat bump patch
```

/cc @zendesk/quokka


sandlerr:
tests please

svizzari:
Nice cleanup and implementation though.  Do we want, or care, about (optionally) git tagging too (if the app is a repo)?

princemaple:
@svizzari I think we care, but that might be a little out of the scope of this PR. We will have to deal with all the complexity about whether it's a repo, whether we have access to git binary, etc. We do have to choice to add another gem...

svizzari:
@princemaple Sure, let's think about that in another PR.  Def, use a gem if one is available+suitable rather than reinventing the wheel.

princemaple:
About reinventing the wheel, the reason I implemented a tiny semver parsing / formating thingy was, I looked through several semver gems, none of them really support parsing non-standard semver string. (i.e. most of our apps have version specified as '1.0' instead of '1.0.0')
Plus most of the gems have all sorts of quirks...

princemaple:
:haircut: @svizzari @sandlerr

princemaple:
:haircut:

spec/lib/zendesk_apps_tools/bump.rb
```diff
@@ -0,0 +1,70 @@
...
+      subject.instance_variable_set(:@manifest, { 'version' => '1.0' })
+    end
+
+    it 'works with incomplete semver version' do
+      VERSION_PARTS.each do |part|
+        expect { subject.send(part) }.not_to raise_error
+      end
+    end
+
+    it 'brings the version to semver - patch' do
```
svizzari: > ensures there is a valid patch version

?

svizzari:
:+1:

lib/zendesk_apps_tools/bump.rb
```diff
@@ -0,0 +1,29 @@
+require 'thor'
+require 'json'
+
+require 'zendesk_apps_tools/manifest_handler'
+
+module ZendeskAppsTools
+  class Bump < Thor
+    include Thor::Actions
+    prepend ManifestHandler
```
adammw: Have not seen this before, I found it rather confusing, but seems to make sense now you've explained it to me.

lib/zendesk_apps_tools/manifest_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+        semver[part] = (semver[part] || '0').to_i
+      end
+    end
+
+    def update_version
+      @manifest['version'] = version
+    end
+
+    def write_manifest
+      File.open('manifest.json', 'w') do |f|
```
atroche: can you extract manifest.json into a constant?

adammw:
:+1:

atroche:
:+1:

PR #82, opened 2015-02-27
ZD#1001781 Declare VERSION_PARTS using backwards compatible format

jcheatham:
@princemaple  looks like https://github.com/zendesk/zendesk_apps_tools/pull/81 broke compatibility with ruby 1.9 per https://support.zendesk.com/agent/tickets/1001781 & https://support.zendesk.com/hc/en-us/articles/203691376#comment_202710433
Not sure if this is enough to fix it, but figured it was a good starting point :)

### Risks
None


princemaple:
Hi @jcheatham, I also used prepend which is also ruby 2+
Ruby 1.9 is no longer getting any updates, even bug fixes, I thought it's time to move on.
Maybe I should just set proper ruby version in the gemspec.

jcheatham:
Hmmm, yeah, specifying the ruby version might be nice for any customers that are locked in to 1.9

princemaple:
Yeah. Sorry fir the inconvenience, I will make the change asap.

princemaple:
Hi, I've released `1.22.2` which restricts Ruby >= 2.0, I've also removed `1.22.0` and `1.22.1`, please see the ticket for more information.

Thanks for the help!

jcheatham:
No problem, thanks!

PR #83, opened 2015-03-28
Patch array to_h for Ruby 2.0

princemaple:
:koala: 

Didn't realize that `Array#to_h` was only added in Ruby 2.1

/cc @zendesk/quokka  @clong-

### Risks
- N/A


princemaple:
Also manually tested with Ruby 2.0.0-p643

itirkarp:
I'd prefer using something else instead of to_h so that we don't have to patch. But if we really need to_h then :+1:

sandlerr:
could use https://github.com/marcandre/backports/blob/master/lib/backports/2.1.0/enumerable/to_h.rb as well.

Why were the tests still green since older versions are in .travis.yml?

princemaple:
@sandlerr because I was stupid and did not name the specs correctly.
RSpec apparently looks for ruby files named with a `_spec` suffix, and I forgot about that.

princemaple:
@sandlerr I mimicked the `Array#to_h` implementation. The gem you linked has a different error message when types don't match up. Also Ruby has a different implementation for `Array#to_h` and `Enumerable#to_h`

sandlerr:
:+1:

princemaple:
For future reference, if we want to use more methods that only exist in 2.1+, we should consider using the gem @sandlerr linked.

princemaple:
Until 2.0 support is dropped, of course.
ruby_2_0_0-p643_released

PR #86, opened 2015-04-27
pass locale to ZAS Package

craig-day:
Use the locale parameter sent by the classic apps helper. 

/cc @pdeuter @benhass @princemaple 

### References
 - Classic Change: https://github.com/zendesk/zendesk/pull/16964

### Risks
 - Low. Just adding a key into the parameters hash.

### DIFF
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -17,9 +17,8 @@ module ZendeskAppsTools
           last_mtime = curr_mtime
         end
       end
-      settings.parameters.merge!(locale: params['locale'])
 
-      ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters)
+      ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters, params['locale'])
     end
   end
 end
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -17,13 +17,9 @@ module ZendeskAppsTools
           last_mtime = curr_mtime
         end
       end
-      settings.parameters.merge!( { locale: normalize_locale(params['locale']) } )
+      settings.parameters.merge!(locale: params['locale'])
 
       ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters)
-    end
-
-    def normalize_locale(locale)
-      locale.downcase.gsub("-us", "")
     end
   end
 end
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -17,8 +17,13 @@ module ZendeskAppsTools
           last_mtime = curr_mtime
         end
       end
+      settings.parameters.merge!( { locale: normalize_locale(params['locale']) } )
 
       ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters)
+    end
+
+    def normalize_locale(locale)
+      locale.downcase.gsub("-us", "")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_tools/server.rb
```diff
@@ -17,8 +17,13 @@ class Server < Sinatra::Base
           last_mtime = curr_mtime
         end
       end
+      settings.parameters.merge!( { locale: normalize_locale(params['locale']) } )
```
princemaple:
```ruby
settings.parameters.merge!(locale: normalize_locale(params['locale']))
```

princemaple:
:+1: otherwise

lib/zendesk_apps_tools/server.rb
```diff
@@ -17,8 +17,13 @@ class Server < Sinatra::Base
...
 
       ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters)
     end
+
+    def normalize_locale(locale)
```
liulikun: `locale` could be nil? Can set default here `def normalize_locale(locale = 'en')`.

liulikun:
:+1:

craig-day:
@princemaple @liulikun Decided to not touch the locale in here and just pass `params['locale']` to ZAS.  I am going to change the logic there a little bit to handle the situation more appropriately.

craig-day:
hey @princemaple what is the version bumping process for this? I would like to bump before and after this merge so its easy to not use if problems come up.

lib/zendesk_apps_tools/server.rb
```diff
@@ -17,6 +17,7 @@ class Server < Sinatra::Base
           last_mtime = curr_mtime
         end
       end
+      settings.parameters.merge!(locale: params['locale'])
```
liulikun: An edge case would be if the app also names a parameter 'locale', then we're going to overwrite it here. Can we pass it through without touching parameters?

princemaple:
Why would you need to bump twice?

craig-day:
it was never bumped after the pseudo translate code was added in. I don't really need to, just a though.

lib/zendesk_apps_tools/server.rb
craig-day: @liulikun we could add an optional parameter to the end of the method signature for `ZendeskAppsSupport::Package#initialize` if that seems better?

princemaple:
I see. Not sure if you have push rights.
I normally just do
```
rake bump:patch
rake release
```

princemaple:
If you don't have push rights, I can do it for you.

lib/zendesk_apps_tools/server.rb
pdeuter: Shouldn't locale be a standard parameter for *all* apps.  I think it should be.  I can hardly imagine that an app would have a locale parameter that did not have the same semantics as this locale parameter.

craig-day: @pdeuter I think he is pointing out the app parameter `locale` could potentially be overwritten by the URL parameter being sent along from a classic helper.  I'm not sure if this is the same issue you are referring to.

lib/zendesk_apps_tools/server.rb
```diff
@@ -18,7 +18,7 @@ class Server < Sinatra::Base
         end
       end
 
-      ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters)
+      ZendeskAppsSupport::Package.new(settings.root).readified_js(nil, settings.app_id, "http://localhost:#{settings.port}/", settings.parameters, params['locale'])
```
craig-day: @princemaple @liulikun how do we feel about doing it this way instead? Corresponding change is [here](https://github.com/zendesk/zendesk_apps_support/pull/83/files#diff-f9dc5da4f07a7cd59a608b6da769f0ecR97)

liulikun: I like this approach better.

craig-day:
@princemaple @liulikun bump.

liulikun:
:+1:

craig-day:
@princemaple @liulikun merged and bumped. Can someone release the new version? Thanks!

princemaple:
Released

PR #90, opened 2015-06-04
Add the ability to commit and tag after bump

princemaple:
:koala: 

It delegates to shell command execution. The reasons I think it's OK to do so:

1. this is opt in
2. this is not on server side

cc @zendesk/quokka

### DIFF
--- a/lib/zendesk_apps_tools/bump.rb
+++ b/lib/zendesk_apps_tools/bump.rb
@@ -8,7 +8,14 @@ module ZendeskAppsTools
     include Thor::Actions
     prepend ManifestHandler
 
+    SHARED_OPTIONS = {
+      ['commit', '-c'] => false,
+      ['message', '-m'] => nil,
+      ['tag', '-t'] => false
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_tools/bump.rb
```diff
@@ -16,16 +23,38 @@ def major
...
+
+    private
+
+    def post_actions
+      return tag if options[:tag]
+      commit if options[:commit]
+    end
+
+    def commit
+      `git commit -am #{commit_message}`
```
itirkarp: This will not add untracked / new files. Will that be a problem?

princemaple: Nope, it's meant to commit the change on manifest.

itirkarp:
Looks ok. Hope it works :+1:

princemaple:
It does, I have been using it already.

tobico:
:+1:

PR #94, opened 2015-07-29
mark the app name as a hidden translation

craig-day:
Mark the app name as a hidden translation. This means translators won't get sent this string for translation if it gets swept by the i18n team. 

/cc @princemaple @liulikun  @zendesk/i18n 

### Risks
 - None


princemaple:
The test failed in a fantastic way. I can work on your branch to fix it if you don't have time.

craig-day:
Actually this might not be needed/wanted by the i18n team. It is a tough one.  Some times we want the app name translated like 'Time Tracking' is something nice that can be translated.  Closing for now.

ydrain:
i18n team will mark `hidden` as needed during the string review. Thanks, @craig-day!

PR #96, opened 2015-09-21
bundle update zendesk_apps_support

ocke:
/cc @zendesk/quokka

### References
 - Jira link: 

### Risks
 - None

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -30,7 +30,7 @@ GEM
       multi_test (>= 0.1.1)
     diff-lcs (1.2.5)
     erubis (2.7.0)
-    execjs (2.5.2)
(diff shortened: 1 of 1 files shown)


xuebinglee:
:+1: on :green_heart: 
These are minor gem upgrades and thus should be fine.

sandlerr:
in fact updating this file does nothing outside of development - if this gem actually depends on any of the updated versions you'll need to specify the dependencies in the gemspec.

ocke:
wut. what? So there is a Gemfile, a Gemfile.lock. But they don't actually do anything...

If I ever create a new language and want to make the dependency ecosystem as complex as possible, I have Ruby as a good starting point.

PR #98, opened 2015-09-28
AF-154 add doc

sandlerr:
:koala:

/cc @zendesk/quokka @chucknado 

### References
 - Jira link: https://zendesk.atlassian.net/browse/AF-154

### DIFF
--- a/doc/tools.md
+++ b/doc/tools.md
@@ -54,6 +54,9 @@ Follow these steps to run an app locally:
 
 To stop the server, switch to your command-line interface and press Control+C.
 
+#### App Settings
+
+In testing some apps it is necessary to specify the settings for the app. `zat server` will ask interactively for the values of all settings specified in the app's `manifest.json`. However, `zat server -c [$CONFIG_FILE]` allows specifying the settings using a JSON or YAML file where the keys are the same as in the manifest. The default filename is `settings.yml`.
 
 ### Package
 
--- a/doc/tools.md
+++ b/doc/tools.md
@@ -56,9 +56,12 @@ To stop the server, switch to your command-line interface and press Control+C.
 
 #### App Settings
 
-In testing some apps it is necessary to specify the settings for the app. `zat server` will ask interactively for the values of all settings specified in the app's `manifest.json`. However, `zat server -c [$CONFIG_FILE]` allows specifying the settings using a JSON or YAML file where the keys are the same as in the manifest. The default filename is `settings.yml`.
+When testing your app, you might need to specify some [app settings](manifest#app-settings). If you started the local server with `zat server`, ZAT will ask interactively for the values of all the settings specified in the app's `manifest.json` file. However, you might prefer to specify the settings in a JSON or YAML file.
 
-As a concrete example if `manifest.json` contains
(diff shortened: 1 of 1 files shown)


pmgrove:
@sandlerr are we able to see an example of how the settings within the manifest are shown in the settings.yml file?

danielbreves:
Why do we need the -c option? Why not always use that if present?

sandlerr:
so you don't have to delete / edit the file when you want to run a one-off config? I don't feel strongly about it, happy to change if it you like.

danielbreves:
I don't feel strongly either, it's only a couple of characters :)

danielbreves:
:+1:

doc/tools.md
```diff
@@ -54,6 +54,39 @@ Follow these steps to run an app locally:
 
 To stop the server, switch to your command-line interface and press Control+C.
 
+#### App Settings
+
+In testing some apps it is necessary to specify the settings for the app. `zat server` will ask interactively for the values of all settings specified in the app's `manifest.json`. However, `zat server -c [$CONFIG_FILE]` allows specifying the settings using a JSON or YAML file where the keys are the same as in the manifest. The default filename is `settings.yml`.
```
chucknado: Suggested edit for the first paragraph: 

> When testing your app, you might need to specify some [app settings](./manifest#app-settings). If you started the local server with  `zat server`, the app will ask interactively for the values of all the settings specified in the app's `manifest.json` file. However, you might prefer to specify the settings in a JSON or YAML file. 

> 1. Create a JSON or YAML file with your settings. The keys in the file should be the same as in your manifest file.
> 2. Start the server with `zat server -c [$CONFIG_FILE]`, where $CONFIG_FILE is the name of your JSON or YAML file. The default filename is `settings.yml`. You don't need to specify it if you use it.

doc/tools.md
```diff
@@ -54,6 +54,39 @@ Follow these steps to run an app locally:
...
+
+As a concrete example if `manifest.json` contains
```
chucknado: Suggested edit:

> ##### Example
> If `manifest.json` contains the following settings:

doc/tools.md
```diff
@@ -54,6 +54,39 @@ Follow these steps to run an app locally:
...
+```json
+...
+"parameters": [
+  {
+    "name": "mySetting"
+  },
+...
+```
+
+There may be settings files like
```
chucknado: Original:

> There may be settings files like

Suggested edit:

> Then you can specify the settings in a file as follows:

doc/tools.md
```diff
@@ -54,6 +54,39 @@ Follow these steps to run an app locally:
...
+```
+
+or
+
+```yaml
+./settings.yml
+mySetting: test value
+```
+
+These may be used by either `zat server -c settings.json` or `zat server -c`.
```
chucknado: Original: 

> These may be used by either `zat server -c settings.json` or `zat server -c`.

Suggested edit:

> With the first file, you'd start the server with `zat server -c settings.json`. With the second file, you'd start it with `zat server -c`.

doc/tools.md
sandlerr: instead of `the app will ask interactively` I put `ZAT will ask interactively`

PR #101, opened 2015-10-26
Disable frame_options protection

danielbreves:
:koala:

This allows developers to include an iframe from the assets folder when using ZAT.

/cc @zendesk/quokka @fandoweb 

### Risks
 - None


sandlerr:
:+1: sinatra implicit opt-in protections are a bit of a nuisance sometimes.

svizzari:
ha, so it was kinda defaulted :+1:

PR #105, opened 2015-11-30
Multiple apps for ZAT

ocke:
:koala:
Multiple Apps for ZAT

It used to be olaf/multiple_apps but it got so many changes that I merged them all into 1 nice commit based on the new master with ZAS 1.20

/cc @zendesk/quokka

### References
 - Jira link: 

### Risks
 - None


lib/zendesk_apps_tools/command.rb
```diff
@@ -90,33 +90,48 @@ def clean
...
+      if app_paths.length != 0 && options[:path] != DEFAULT_SERVER_PATH
+        say_error_and_exit "please either use -p or list the directory structure directly"
+      end
 
-      settings_helper = ZendeskAppsTools::Settings.new
+      if app_paths.length != 0 && options[:config] != DEFAULT_CONFIG_PATH
+        say_error_and_exit "cannot use -c in combination with multiple apps"
+      end
+
+      if app_paths.empty?
```
adammw: Why not use empty? in the conditons above or change this to check length == 0

lib/zendesk_apps_tools/command.rb
```diff
@@ -90,33 +90,48 @@ def clean
...
+        app_paths << options[:path]
+      end
 
-      settings = settings_helper.get_settings_from_file options[:config], manifest['parameters']
+      apps = app_paths.map do | path |
```
adammw: I'm tempted to suggest moving this code into the server.rb file and just pass the app_paths in as a setting.

lib/zendesk_apps_tools/command.rb
```diff
@@ -90,33 +90,48 @@ def clean
...
+        {
+          package: package,
+          settings_file_path: settings_file_path,
+          settings: settings
+        }
+      end
 
-      unless settings
-        settings = settings_helper.get_settings_from_user_input self, manifest['parameters']
+      if apps.empty?
```
adammw: This would never be hit, as you already check app_paths.empty? above and make it non-empty (with the addition of the path)

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,68 @@
 require 'sinatra/base'
 require 'zendesk_apps_support/package'
+require 'zendesk_apps_support/installed'
+include ZendeskAppsTools::PackageHelper
```
adammw: where is this used?

ocke: no idea. I moved some stuff around so it can go

ocke:
@zendesk/quokka comments?

lib/zendesk_apps_tools/command.rb
```diff
@@ -90,33 +90,42 @@ def clean
     DEFAULT_SERVER_PATH = './'
     DEFAULT_CONFIG_PATH = './settings.yml'
     DEFAULT_SERVER_PORT = 4567
-    DEFAULT_APP_ID = 0
 
     desc 'server', 'Run a http server to serve the local app'
     method_option :path, default: DEFAULT_SERVER_PATH, required: false, aliases: '-p'
     method_option :config, default: DEFAULT_CONFIG_PATH, required: false, aliases: '-c'
     method_option :port, default: DEFAULT_SERVER_PORT, required: false
-    method_option :app_id, default: DEFAULT_APP_ID, required: false
```
adammw: We're no longer providing this functionality? I found it quite useful for live debugging

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,83 @@
...
-      )
-
-      ZendeskAppsSupport::Installed.new([app_js], [installation]).compile_js
+      installed = ZendeskAppsSupport::Installed.new(appsjs, installations)
+      installed.compile_js(installation_orders: order)
+    end
+
+    get "/:file" do |file|
+      response.status = 404
+      'Deprecated. We support multiple apps now, please add app_id to the path for file, e.g. /0/' +file
```
adammw: Reflection of input usually brings problems, do we really need it?

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,83 @@
...
+    end
+
+    get "/:app_id/:file" do |app_id, file|
+      # convert to postive and -1. So -1 => 0, -3 => 2, etc
+      id = (-app_id.to_i)-1
```
adammw: `index` perhaps would make more sense than id

lib/zendesk_apps_tools/settings.rb
```diff
@@ -57,8 +71,8 @@ def get_settings_from_file(filepath, parameters)
         end
 
         if !input && param['required']
-          puts "'#{param['name']}' is required but not specified in the config file.\n"
-          return nil
+          puts "\e[0;31m'#{param['name']}' is required but not specified in the config file.\e[0m\n"
```
adammw: don't we have a library that can abstract these escape codes?

lib/zendesk_apps_tools/command.rb
adammw: Perhaps we could have it as the starting index now? (defaulting to -1)

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,83 @@
...
+          locations = [ locations ] if locations.is_a?(String)
+          locations.each do |location, index|
+            order[location] ||= []
+            order[location] << app_id
+          end
+        end
+
+        if app[:settings_file_path]
+          curr_mtime = File.stat(app[:settings_file_path]).mtime
+          curr_domain = params['subdomain']
```
adammw: Why do you care about the subdomain (and where is this even set)?

lib/zendesk_apps_tools/command.rb
ocke: We can make it a one-app-only option. If you load more than 1 app the option will do nothing. Because we don't know to WHICH app you want to apply this, and what will be the number for the next app.

How about that?

lib/zendesk_apps_tools/settings.rb
ocke: Well I couldn't get it to work. I'll have another try.

lib/zendesk_apps_tools/server.rb
ocke: it's a parameter! app.js?locale=en&subdomain=z3nzibar.

I added this a while ago to the framework.

ocke: It's totally a dingleberry btw.

sandlerr:
this PR is a hydra! Cut off one HEAD and two others take its place.

lib/zendesk_apps_tools/command.rb
sandlerr: yes, make it work for one app

lib/zendesk_apps_tools/command.rb
```diff
@@ -90,33 +90,47 @@ def clean
...
-    def server
-      setup_path(options[:path])
-      manifest = app_package.manifest_json
+    method_option :app_id, required: false
+    def server(*app_paths)
+      if !app_paths.empty? && options[:path] != DEFAULT_SERVER_PATH
```
sandlerr: pull out the repeated condition?

lib/zendesk_apps_tools/settings.rb
```diff
@@ -25,7 +26,18 @@ def get_settings_from_user_input(user_input, parameters)
       end
     end
 
-    def get_settings_from_file(filepath, parameters)
+    def find_settings_file(path)
+      try_paths = ['./', '../', 'settings/', '../settings/']
```
sandlerr: I think this is too many default options

lib/zendesk_apps_tools/settings.rb
```diff
@@ -25,7 +26,18 @@ def get_settings_from_user_input(user_input, parameters)
...
+
+      try_paths.reduce([]) do | memo, try_path |
+        memo << File.join(path, try_path, 'settings.yml')
+        memo << File.join(path, try_path, 'settings.json')
+      end.find do |settings_file|
+        File.exists?(settings_file)
+      end
+    end
+
+    def get_settings_from_file(filepath, parameters, domain = false)
```
sandlerr: domain is not really a boolean, let's default to `nil`

lib/zendesk_apps_tools/settings.rb
```diff
@@ -38,6 +50,8 @@ def get_settings_from_file(filepath, parameters)
           settings_data = YAML.load(settings_file)
         end
 
+        settings_data = settings_data[domain] if settings_data[domain] && domain
```
sandlerr: why the `&& domain`?

lib/zendesk_apps_tools/settings.rb
sandlerr: if not, you could set them to constants / local variables like we have [here](https://github.com/zendesk/docker-images/blob/master/dotfiles/.bashrc#L18-L26)

spec/lib/zendesk_apps_tools/settings_spec.rb
```diff
@@ -182,7 +183,7 @@
         expect(@context.get_settings_from_file('spec/fixture/config/settings.yml', parameters)).to eq(settings)
       end
 
-      it 'returns nil because you forgot to specifiy a required field without a default' do
+      it 'sets nil because you forgot to specifiy a required field without a default' do
```
sandlerr: `specify` has an extra i

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,78 @@
 require 'sinatra/base'
 require 'zendesk_apps_support/package'
```
sandlerr: is this still used?

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,78 @@
...
+      settings_helper = ZendeskAppsTools::Settings.new
+
+      appsjs = []
+      installations = []
+      order = {}
+
+      settings.apps.each_with_index do |app, index|
+        package = app[:package]
+        app_id = installation_id = -(index+1)
+        app_name = package.manifest_json['name'] || 'Local App'
```
sandlerr: does this break things if you have two nameless apps?

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,78 @@
...
+
+        appsjs << package.compile_js(
+          app_name: app_name,
+          app_id: app_id,
+          assets_dir: "http://localhost:#{settings.port}/#{app_id}/",
+          locale: params['locale']
+        )
+
+        location = begin
```
sandlerr: this isn't a great variable name

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,78 @@
...
+          locations = package.manifest_json['location']
+          locations = [ locations ] if locations.is_a?(String)
+          locations.is_a?(Hash) ? locations : { zendesk: locations }
+        end
+
+        location.each do |host, locations|
+          locations = [ locations ] if locations.is_a?(String)
+          locations.each do |location, index|
```
sandlerr: not using index

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,45 +1,78 @@
...
+require 'zendesk_apps_support/installed'
 
 module ZendeskAppsTools
   class Server < Sinatra::Base
     set :protection, :except => :frame_options
-    set :public_folder, proc { "#{settings.root}/assets" }
     last_mtime = Time.new(0)
+    last_domain = nil
 
     get '/app.js' do
```
sandlerr: this method is really long, could you refactor it to shorter ones?

lib/zendesk_apps_tools/command.rb
```diff
@@ -90,33 +90,47 @@ def clean
...
+        say_error_and_exit "please either use -p or list the directory structure directly"
```
sandlerr: > list the directory structure directly

what does this mean?

sandlerr:
can you add some docs? or at least commandline help?

lib/zendesk_apps_tools/command.rb
```diff
@@ -115,33 +115,47 @@ def clean
...
 
     desc 'server', 'Run a http server to serve the local app'
     method_option :path, default: DEFAULT_SERVER_PATH, required: false, aliases: '-p'
     method_option :config, default: DEFAULT_CONFIG_PATH, required: false, aliases: '-c'
     method_option :port, default: DEFAULT_SERVER_PORT, required: false
-    method_option :app_id, default: DEFAULT_APP_ID, required: false
-    def server
-      setup_path(options[:path])
-      manifest = app_package.manifest_json
+    method_option :app_id, required: false
```
danielbreves: The notify API in ZAM checks if the app id is 0 to check if we're running zat

lib/zendesk_apps_tools/settings.rb
```diff
@@ -25,7 +26,16 @@ def get_settings_from_user_input(user_input, parameters)
       end
     end
 
-    def get_settings_from_file(filepath, parameters)
+    def find_settings_file(path)
+      ['./settings.yml', './settings.json'].reduce([]) do | memo, try_file |
+        memo << File.join(path, try_file)
+        memo << File.join(path, try_file)
```
danielbreves: Why do you need to do this twice?

lib/zendesk_apps_tools/server.rb
```diff
@@ -1,47 +1,65 @@
...
+        app_name = package.manifest_json['name'] || 'Local App'
+
+        appsjs << package.compile_js(
+          app_name: app_name,
+          app_id: app_id,
+          assets_dir: "http://localhost:#{settings.port}/#{app_id}/",
+          locale: params['locale']
+        )
+
+        if app[:settings_file_path]
```
danielbreves: Can you extract this to a separate method and perhaps add a comment explaining what it's doing

spec/spec_helper.rb
```diff
@@ -10,6 +10,8 @@
 $LOAD_PATH.unshift(File.dirname(__FILE__) + '/../lib')
 $LOAD_PATH.unshift(File.dirname(__FILE__) + '/../lib/zendesk_apps_tools')
 
+ENV['ZAT_DEV'] = 'true'
```
danielbreves: `ENV["ZAT_ENV"] = 'test'`?

PR #107, opened 2015-12-15
Bump sinatra

ocke:
:koala:

a bug in sinatra, so we need to bump the version.

/cc @zendesk/quokka

### References
 - Jira link: 

### Risks
 - None


sandlerr:
something to do with routing?
:+1: on green

ocke:
There was a bug in the communication of sinatra and rack. I bumped the sinatra version so it works with our rack version again.

danielbreves:
Any breaking changes from 1.3 to 1.4?

danielbreves:
:+1:

ocke:
not as far as I know, I tested it and the error page works again (like it used to). App.js loads like it should too (and logo.png)

PR #109, opened 2016-01-06
Access-Control-Allow-Origin: *

danielbreves:
:koala:

Seems like we need this since the last [sinatra bump](https://github.com/zendesk/zendesk_apps_tools/pull/107)

/cc @zendesk/quokka

### Risks
 - None


adammw:
Perhaps we should do this as middleware for all URLs? Assets would have the same problem too...

timbertson:
@adammw I think we would only need it for assets if they're loaded via AJAX (seems unlikely, but I've no idea). :+1:

ocke:
:+1:

PR #110, opened 2016-01-06
Stricter access control

danielbreves:
:koala:

/cc @zendesk/quokka

### Risks
 - None

### DIFF
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -6,9 +6,10 @@ module ZendeskAppsTools
     set :protection, :except => :frame_options
     set :public_folder, proc { "#{settings.root}/assets" }
     last_mtime = Time.new(0)
+    ZENDESK_DOMAINS_REGEX = /^http(?:s)?:\/\/[a-z0-9-]+\.(?:zendesk|zopim|zd-(?:dev|master|staging))\.com$/
 
     get '/app.js' do
-      headers 'Access-Control-Allow-Origin' => '*'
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_tools/server.rb
```diff
@@ -6,9 +6,10 @@ class Server < Sinatra::Base
     set :protection, :except => :frame_options
     set :public_folder, proc { "#{settings.root}/assets" }
     last_mtime = Time.new(0)
+    ZENDESK_DOMAINS_REGEX = /^http(?:s)?:\/\/[a-z0-9-]+\.(?:zendesk|zopim|zd-(?:dev|master|staging))\.com$/
 
     get '/app.js' do
-      headers 'Access-Control-Allow-Origin' => '*'
+      headers 'Access-Control-Allow-Origin' => '*' if valid_origin
```
adammw: `=request.env['HTTP_ORIGIN']`?

danielbreves:
@zendesk/quokka :haircut:

timbertson:
:+1:

sandlerr:
setting it to the origin (instead of *) will send the session over, do you care? Shouldn't matter since its the local machine but also should be useless since it is encrypted.

danielbreves:
@adammw any security concerns with leaving it '*'?

adammw:
:+1:

timbertson:
@danielbreves only thing I can think of is that the browser _might_ cache the "*" value and use it to allow future requests when served from cache. But mostly it just seems a bit improper.

PR #116, opened 2016-02-19
Make app template use the default i18n format as documented

danielbreves:
:bread: 

/cc @zendesk/quokka

### Description
Make app template use the default i18n format as documented [here](https://developer.zendesk.com/apps/docs/agent/i18n).

### References
* HC community: https://support.zendesk.com/hc/en-us/community/posts/209137097-Whats-the-purpose-of-the-package-app-that-messes-with-translations- 

### Risks
* [low] Translations may not work for new apps created using ZAT

### DIFF
--- a/app_template/translations/en.json
+++ b/app_template/translations/en.json
@@ -1,54 +1,16 @@
 {
   "app": {
-    "package": "app_name",
-    "description": {
-      "value": "Play the famous zen tunes in your help desk.",
-      "title": "app description"
-    },
-    "name": {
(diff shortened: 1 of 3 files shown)


sandlerr:
:+1:
should we create a zendesk-specific option for `zat new` to output the non-default?

danielbreves:
@sandlerr https://zendesk.atlassian.net/browse/AF-254

svizzari:
:+1:

adammw:
:+1:

PR #119, opened 2016-04-20
Why wait, life is too short

ocke:
:v:

/cc @zendesk/vegemite
/cc @catkins 

### Description

Threaded translation.

Like a 100 times faster

Ok 9 times faster. But still... it feels a lot faster
### References
* JIRA: 

### Risks
- Too much time on your hands

### DIFF
--- a/lib/zendesk_apps_tools/translate.rb
+++ b/lib/zendesk_apps_tools/translate.rb
@@ -49,14 +49,14 @@ module ZendeskAppsTools
 
     desc 'update', 'Update translation files from Zendesk'
     method_option :path, default: './', required: false
-    def update(request_builder = Faraday.new)
+    def update()
       setup_path(options[:path]) if options[:path]
       app_package = get_value_from_stdin('What is the package name for this app? (without app_)', valid_regex: /^[a-z_]+$/, error_msg: 'Invalid package name, try again:')
 
       key_prefix = "txt.apps.#{app_package}."
 
       say('Fetching translations...')
-      locale_response = api_request(LOCALE_ENDPOINT, request_builder)
+      locale_response = Faraday.get(LOCALE_ENDPOINT)
 
       if locale_response.status == 200
         locales = JSON.parse(locale_response.body)['locales']
--- a/lib/zendesk_apps_tools/translate.rb
+++ b/lib/zendesk_apps_tools/translate.rb
@@ -61,10 +61,10 @@ module ZendeskAppsTools
       if locale_response.status == 200
         locales = JSON.parse(locale_response.body)['locales']
 
+        locales = locales.map { |locale| fetch_locale_async locale, app_package}.map(&:value)
+
         locales.each do |locale|
-          locale_url      = "#{locale['url']}?include=translations&packages=app_#{app_package}"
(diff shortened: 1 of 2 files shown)


catkins:
:+1::ok_hand::100: love that copypasta!

(FWIW @ocke lifted this technique from our repo, and it's been working a treat)

lib/zendesk_apps_tools/translate.rb
```diff
@@ -95,6 +95,14 @@ def self.source_root
     end
 
     no_commands do
+      def fetch_locale_async(locale, app_package)
+        Thread.new do
+          say("Fetching #{locale['locale']}")
```
catkins: If the output gets minced by the multithreading, you might have to set `$stdout = true` at the start of the command.

svizzari: Nice work guys!

ocke: yeah I was talking to chris and he showed me his technique. So I implemented this.
* Old translate on answer suggestion app: 45.252s
* Fast translate on answer suggestion app: 5.444s

ocke: @catkins The output looks good btw

catkins: Ah cool. Thor must sync the output somewhere.

sandlerr:
no idea how this works, don't you have to call `join` on the threads to block on their output somewhere?

catkins:
@sandlerr Invoking `#value` does a `#join`

http://ruby-doc.org/core-2.2.0/Thread.html#method-i-value

catkins:
@sandlerr the `.map(&:value)` waits until all of the tasks are done

```
locales.map { |locale| fetch_locale_async locale, app_package}.map(&:value)
```

lib/zendesk_apps_tools/translate.rb
```diff
@@ -49,14 +49,14 @@ def to_json
...
-    def update(request_builder = Faraday.new)
+    def update()
       setup_path(options[:path]) if options[:path]
       app_package = get_value_from_stdin('What is the package name for this app? (without app_)', valid_regex: /^[a-z_]+$/, error_msg: 'Invalid package name, try again:')
 
       key_prefix = "txt.apps.#{app_package}."
 
       say('Fetching translations...')
-      locale_response = api_request(LOCALE_ENDPOINT, request_builder)
+      locale_response = Faraday.get(LOCALE_ENDPOINT)
```
svizzari: I think the use of `request_builder` was so that `update` could be stubbed more easily in tests, is that no longer true?

ocke: I stubbed it in a different way (see spec/lib/zendesk_apps_tools/translate_spec.rb)

svizzari: Yea, I noticed the updated spec after commenting, and then didn't come back to comment again :/

ocke: so where's my 👍 ? ;-)

svizzari:
👍

PR #120, opened 2016-04-22
replace an unhandled crash with a handled crash

sandlerr:
@zendesk/vegemite bring our handling of invalid json when polling the job status in line with the way we handle it everywhere else

### DIFF
--- a/lib/zendesk_apps_tools/deploy.rb
+++ b/lib/zendesk_apps_tools/deploy.rb
@@ -13,9 +13,7 @@ module ZendeskAppsTools
 
       check_status response
 
-    rescue Faraday::Error::ClientError => e
-      say_error_and_exit e.message
-    rescue JSON::ParserError => e
+    rescue Faraday::Error::ClientError, JSON::ParserError => e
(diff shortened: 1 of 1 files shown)


danielbreves:
👍

ocke:
👍

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -8,19 +8,20 @@ def deploy_app(connection_method, url, body)
...
         req.headers[:content_type] = 'application/json'
-
         req.body = JSON.generate body
       end
 
       check_status response
 
     rescue Faraday::Error::ClientError => e
       say_error_and_exit e.message
+    rescue JSON::ParserError => e
```
svizzari: I believe:
```ruby
rescue Faraday::Error::ClientError, JSON::ParserError => e
  say_error_and_exit e.message
end
```

would be more concise, but may invalidate the reason for this PR?

sandlerr:
@svizzari refactored

ocke:
👍

PR #121, opened 2016-04-22
pretty colours and exit code

sandlerr:
@zendesk/vegemite @zendesk/wombat


sandlerr:
since we are relying on this for deploys we really should `exit 1` when things go wrong

svizzari:
👍 on 🍏

danielbreves:
👍

PR #127, opened 2016-05-17
Satisfy preflight requests

ocke:
:v:

/cc @zendesk/vegemite

### Description
Browsers turn out to be draconian when it comes to getting anything else than an images or a javascript file.

Turns out 2 headers were added which makes this a CORS request with a preflight where we need to allow those headers

This PR satisfies all that, so i can load a .txt or .json file from my assets folder.

### References
* JIRA: 

### Risks

### DIFF
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -4,7 +4,6 @@ require 'zendesk_apps_support/package'
 module ZendeskAppsTools
   class Server < Sinatra::Base
     set :protection, :except => :frame_options
-    set :public_folder, proc { "#{settings.root}/assets" }
     last_mtime = Time.new(0)
     ZENDESK_DOMAINS_REGEX = /^http(?:s)?:\/\/[a-z0-9-]+\.(?:zendesk|zopim|zd-(?:dev|master|staging))\.com$/
 
@@ -42,6 +41,18 @@ module ZendeskAppsTools
       )
 
       ZendeskAppsSupport::Installed.new([app_js], [installation]).compile_js
+    end
+
+    get "/:file" do |file|
+      access_control_allow_origin
+      send_file File.join(settings.root, 'assets', file)
+    end
+
+    # This is for any preflight request
+    options "*" do
+      access_control_allow_origin
+      origin = request.env['HTTP_ORIGIN']
+      headers 'Access-Control-Allow-Headers' => request.env['HTTP_ACCESS_CONTROL_REQUEST_HEADERS'] if origin =~ ZENDESK_DOMAINS_REGEX
     end
 
     def access_control_allow_origin
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -49,12 +49,14 @@ module ZendeskAppsTools
     end
 
     # This is for any preflight request
+    # It reads 'Access-Control-Request-Headers' to set 'Access-Control-Allow-Headers'
+    # And also sets 'Access-Control-Allow-Origin' header
     options "*" do
-      #don't delete this
       access_control_allow_origin
       headers 'Access-Control-Allow-Headers' => request.env['HTTP_ACCESS_CONTROL_REQUEST_HEADERS'] if request.env['HTTP_ORIGIN'] =~ ZENDESK_DOMAINS_REGEX
     end
 
+    # This sets the 'Access-Control-Allow-Origin' header for requests coming from zendesk
     def access_control_allow_origin
       origin = request.env['HTTP_ORIGIN']
       headers 'Access-Control-Allow-Origin' => origin if origin =~ ZENDESK_DOMAINS_REGEX
--- a/lib/zendesk_apps_tools/server.rb
+++ b/lib/zendesk_apps_tools/server.rb
@@ -50,9 +50,9 @@ module ZendeskAppsTools
 
     # This is for any preflight request
     options "*" do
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_tools/server.rb
```diff
@@ -44,6 +43,18 @@ class Server < Sinatra::Base
...
+    get "/:file" do |file|
+      access_control_allow_origin
+      send_file File.join(settings.root, 'assets', file)
+    end
+
+    # This is for any preflight request
+    options "*" do
+      access_control_allow_origin
+      origin = request.env['HTTP_ORIGIN']
+      headers 'Access-Control-Allow-Headers' => request.env['HTTP_ACCESS_CONTROL_REQUEST_HEADERS'] if origin =~ ZENDESK_DOMAINS_REGEX
```
sandlerr: why do you need this and the previous line?

lib/zendesk_apps_tools/server.rb
```diff
@@ -44,6 +43,18 @@ class Server < Sinatra::Base
       ZendeskAppsSupport::Installed.new([app_js], [installation]).compile_js
     end
 
+    get "/:file" do |file|
+      access_control_allow_origin
+      send_file File.join(settings.root, 'assets', file)
```
sandlerr: it would probably be best if this didn't support `assets/subdirectory/../file_in_root_dir.png` because that won't work when you upload the app

ocke: noted. what happens with files if i upload something in that order? they get lost? or it gets flattened?

Also can you do ../../../../../etc/hosts with this style?

lib/zendesk_apps_tools/server.rb
ocke: it's a cors request because the framework adds 2 headers to the request. Which are not allowed (without this line).

sandlerr: but doesn't calling `access_control_allow_origin` above already do that?

lib/zendesk_apps_tools/server.rb
sandlerr: it seems like it doesn't do allow you to traverse back up the path. I don't think the filesystem would allow you to create a file called `..` though, @fbvilela might be able to confirm

ocke: with 'it' you mean File.join or send_file?

lib/zendesk_apps_tools/server.rb
ocke: No one is for: Access-Control-Allow-Origin, the preflight need that header and also an additional 'Access-Control-Allow-Headers' header.

ocke: And to make it more confusing it copies the value from yes another header called: Access-Control-Requests-Headers.

sandlerr: can you put a comment so nobody deletes it?

lib/zendesk_apps_tools/server.rb
sandlerr: File.join

sandlerr: At least I wasn't able to do `File.read File.join 'some_dir', '..', 'file_in_parent'`

ocke: ok, so this is good to go?

sandlerr: yep should be fine

ocke:
👍 s anybody???

lib/zendesk_apps_tools/server.rb
```diff
@@ -44,6 +43,18 @@ class Server < Sinatra::Base
...
+    end
+
+    # This is for any preflight request
+    options "*" do
+      #don't delete this
```
sandlerr: Sorry, let's not bother about this. It isn't helpful.

sandlerr:
+1

lib/zendesk_apps_tools/server.rb
```diff
@@ -4,7 +4,6 @@
 module ZendeskAppsTools
   class Server < Sinatra::Base
     set :protection, :except => :frame_options
-    set :public_folder, proc { "#{settings.root}/assets" }
```
svizzari: Wouldn't keeping this and not having your

```ruby
get "/:file"
  ...
end
```

be equivalent?

It's really the `options` block you need, isn't it?

ocke: no. access_control_allow_origin

ocke: I need to run `access_control_allow_origin`

ocke: there's probably some magic you can do to mimic this.

before_public_folder_call.method_send(:access_control_allow_origin) or something, but I don't know magic rails.

sandlerr: much simpler than rails :-) http://www.sinatrarb.com/

svizzari: You could do something like this tho, https://github.com/britg/sinatra-cross_origin

svizzari:
Actually, your implementation is similar, and lighter than the cross_origin gem.  And that gem doesn't handle OPTIONS preflight requests properly, so... 

👍

lib/zendesk_apps_tools/server.rb
ocke: but ... why?

PR #134, opened 2016-07-07
allow serving files from nested asset directories with CORS

sandlerr:
@zendesk/vegemite currently if you have a file like `assets/client/build/index.html` it will 404. Something to do with the `File.join`.

`public_folder` can't be set inside `server.rb` because that is run before we set the `path`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,7 @@ PATH
       faraday (~> 0.9.2)
       rubyzip (~> 0.9.1)
       sinatra (~> 1.4.6)
+      sinatra-cross_origin (~> 0.3.1)
       thor (~> 0.18.1)
       zendesk_apps_support (~> 1.29)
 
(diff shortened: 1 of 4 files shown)


sandlerr:
bug seems to be introduced in https://github.com/zendesk/zendesk_apps_tools/pull/127

adammw:
👍

ocke:
👍

danielbreves:
@sandlerr any way we can test this?

sandlerr:
The best way I can think of would be integration tests that run `zat create --v2` followed by `zat server`, create a file in a subdirectory and then curl it. I can work on that later in the week.

sandlerr:
I mean we could just use http://www.sinatrarb.com/testing.html to make a regression test for this specific bug but I don't think it brings as much value.

ocke:
I tried this with @sandlerr try, and it seems to work 👍

PR #135, opened 2016-07-11
version printing

ocke:
:v:

/cc @zendesk/vegemite

### Description
Print the version when doing `zat -v` or `zat --version`

### References
* JIRA: 

### Risks

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -20,6 +20,8 @@ module ZendeskAppsTools
       clean: false
     }
 
+    map %w[--version -v] => :__print_version
+
     source_root File.expand_path(File.join(File.dirname(__FILE__), '../..'))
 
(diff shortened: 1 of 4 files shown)


adammw:
:+1:

sandlerr:
many gems use a pattern like https://github.com/zendesk/zendesk_api_client_rb/blob/master/lib/zendesk_api/version.rb and reference that constant in the gemspec and anywhere else it is needed. Maybe we can do that?

Also why the double underscore method name?

ocke:
@sandlerr i can use that pattern but does that work with rake bump?

sandlerr:
seems like it: https://github.com/gregorym/bump#supported-locations

ocke:
I removed the underscores, and then it stopped working. I think they mean something special.

sandlerr:
:+1:

PR #137, opened 2016-08-09
support refactors in ZAS

sandlerr:
Support https://github.com/zendesk/zendesk_apps_support/pull/127

Just changes some syntax

cc @zendesk/wombat @zendesk/vegemite 

### Risks
- [low] `zat server` or `zat validate` doesn't work

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
 gemspec
+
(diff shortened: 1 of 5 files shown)


timbertson:
LGTM, assuming you're planning to wait until https://github.com/zendesk/zendesk_apps_support/pull/127 is released before merging 👍

xuebingli:
👍  LGTM

ocke:
Can we please revert the name change! People are complaining and it's very silly all apps are now called 'Local App'.

In fact I want to go the other way, personally I want to remove name from ZAM, it should come from the app! We stick everything in the zip except the name?!?

PR #138, opened 2016-08-29
explain the easiest solution for validation on windows

sandlerr:
`jshintrb` is not compatible with JScript, the JS runtime that ships with windows.
Installing node standalone is easier than any of the other [supported runtimes](https://github.com/rails/execjs/blob/master/lib/execjs/runtimes.rb#L82-L90) due to their compilation toolchains.
cc @zendesk/vegemite


sandlerr:
currently, `zat validate` on windows prints a very sad backtrace

danielbreves:
👍

PR #139, opened 2016-08-30
better error for app not found, color on windows

sandlerr:
@zendesk/vegemite


bin/zat
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
 lib_dir = File.expand_path('../../lib', __FILE__)
 $LOAD_PATH << lib_dir unless $LOAD_PATH.include?(lib_dir)
+ENV['THOR_SHELL'] = 'Color'
```
sandlerr: thor should do this by default - Ruby supports ANSI colours on windows since ruby 2.0 (the lowest we support).

ocke:
👍 In @sandlerr we trust.

PR #141, opened 2016-09-01
fix crash in zat server

sandlerr:
modified this method to no longer need `self` passed in, it's saved in the constructor now.

cc @zendesk/vegemite


danielbreves:
👍

svizzari:
:+1:

PR #142, opened 2016-09-02
fix dev environment

sandlerr:
Oops, my local version of ZAS was built against https://github.com/zendesk/zendesk_apps_support/pull/131 now the Gemfile.lock is broken:

https://travis-ci.org/zendesk/zendesk_apps_tools/jobs/156989834

@zendesk/vegemite 

ZAS 3.1.0 depends on json-stream, 3.1.1 does not.


ocke:
👍

PR #147, opened 2016-10-06
make v2 apps the default

sandlerr:
### Description

Makes the default output of `zat new` to be a v2 app.

@zendesk/vegemite 

#### Passing a value that's already set by default

screen shot 2016-10-06 at 3 56 56 pm

#### Trying to use both, you crazy javascript programmers

screen shot 2016-10-06 at 3 57 06 pm

#### The new default in action

screen shot 2016-10-06 at 3 57 31 pm

#### The v2 option doesn't show up any more even though we haven't deleted it

screen shot 2016-10-06 at 3 57 58 pm

### References

come on @pmgrove what's the story?

### Risks

- [low] `zat new` doesn't work or surprises people

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -28,9 +28,17 @@ module ZendeskAppsTools
     desc 'new', 'Generate a new app'
     method_option :'iframe-only', type: :boolean,
                                   default: false,
-                                  desc: 'Create an iFrame Only app template',
-                                  aliases: ['-i', '--v2']
+                                  hide: true,
+                                  aliases: ['--v2']
+    method_option :v1, type: :boolean, default: false, desc: 'Create a version 1 app template'
     def new
+      if options[:'iframe-only']
+        if options[:v1]
+          say_error_and_exit "Apps can't have both v1 and v2 set."
+        else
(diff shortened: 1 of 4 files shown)


sandlerr:
thoughts on changing the cucumber formatter to `pretty` instead of `progress`?

lib/zendesk_apps_tools/command.rb
```diff
@@ -28,9 +28,17 @@ class Command < Thor
...
-                                  aliases: ['-i', '--v2']
+                                  hide: true,
+                                  aliases: ['--v2']
+    method_option :v1, type: :boolean, default: false, desc: 'Create a version 1 app template'
     def new
+      if options[:'iframe-only']
+        if options[:v1]
+          say_error_and_exit "Apps can't have both v1 and v2 set."
+        else
+          say_status 'warning', 'v2 is the default for new apps, the --v2 option has no effect.', :yellow
```
svizzari: This warning will display if `--iframe-only` is explicitly used? Maybe put:

> v2 is the default for new apps, the --v2 and --iframe-only options have no effect.

?

sandlerr: I'll add that although we never documented --iframe-only afaik

sandlerr:
@svizzari 💇‍♂️

PR #148, opened 2016-10-11
OFFAPPS-691 zat translate update --unattended

sandlerr:
### Description

- Refactor the way we do shared_options because thor's `method_options` doesn't provide as much niceness as `method_option`.
- Implement a `--unattended` option for some commands so that nobody is ever left with a stuck prompt waiting for input.
- Refactor of `zat translate SUBCOMMAND` especially `zat translate update` to be more consistent with linter rules and to `exit 1` instead of `exit 0` on error

cc @zendesk/vegemite @zendesk/wombat 

### References

https://zendesk.atlassian.net/browse/OFFAPPS-691

### Risks

- [low] since this affects options across many commands, many of them might crash / not work

### DIFF
--- a/lib/zendesk_apps_tools/command_helpers.rb
+++ b/lib/zendesk_apps_tools/command_helpers.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'zendesk_apps_tools/common'
 require 'zendesk_apps_tools/api_connection'
 require 'zendesk_apps_tools/deploy'
--- a/lib/zendesk_apps_tools/command_helpers.rb
+++ b/lib/zendesk_apps_tools/command_helpers.rb
@@ -14,6 +14,10 @@ module ZendeskAppsTools
     include ZendeskAppsTools::Directory
     include ZendeskAppsTools::PackageHelper
 
+    def self.included(base)
+      base.extend(ClassMethods)
+    end
+
     def cache
       @cache ||= begin
         require 'zendesk_apps_tools/cache'
--- a/lib/zendesk_apps_tools/common.rb
+++ b/lib/zendesk_apps_tools/common.rb
@@ -1,5 +1,33 @@
+# frozen_string_literal: true
 module ZendeskAppsTools
   module Common
+    module ClassMethods
+      def shared_options(except: [])
+        unless except.include? :path
+          method_option :path,
+                        type: :string,
+                        default: './',
+                        aliases: ['-p']
+        end
+        unless except.include? :clean
+          method_option :clean,
+                        type: :boolean,
+                        default: false
+        end
+        unless except.include? :unattended
+          method_option :unattended,
+                        type: :boolean,
+                        default: false,
+                        desc: 'Experimental: Never prompt for input, expecting all input from the original invocation. Many '\
+                              'commands invoked with this option will just crash.'
+        end
+      end
+    end
+
+    def self.included(base)
+      base.extend(ClassMethods)
+    end
+
     def say_error_and_exit(msg)
       say_error msg
       exit 1
--- a/lib/zendesk_apps_tools/common.rb
+++ b/lib/zendesk_apps_tools/common.rb
@@ -48,11 +48,8 @@ module ZendeskAppsTools
 
       while input = ask(prompt, thor_options)
         return '' if options[:allow_empty] && input.empty?
-        if input.to_s =~ options[:valid_regex]
-          break
-        else
-          say_error options[:error_msg]
-        end
+        break if input.to_s =~ options[:valid_regex]
+        say_error options[:error_msg]
       end
 
       input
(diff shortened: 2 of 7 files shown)


lib/zendesk_apps_tools/common.rb
```diff
@@ -10,6 +41,10 @@ def say_error(msg)
     end
 
     def get_value_from_stdin(prompt, opts = {})
+      if options[:unattended]
```
svizzari: I wonder if `--no-input` or `--without-input` might be slightly more descriptive of the option?

sandlerr: not sure, in the context of translations what is input? the app code? yml or json files? the remote API endpoints we use?

lib/zendesk_apps_tools/command_helpers.rb
```diff
@@ -14,6 +14,10 @@ module CommandHelpers
     include ZendeskAppsTools::Directory
     include ZendeskAppsTools::PackageHelper
 
+    def self.included(base)
```
danielbreves: Isn't this included by Common??

lib/zendesk_apps_tools/common.rb
```diff
@@ -1,5 +1,36 @@
...
+            type: :boolean,
+            default: false
+          }
+        end
+        unless except.include? :unattended
+          method_option :unattended, {
+            type: :boolean,
+            default: false,
+            desc: 'Never prompt for input, expecting all input from the original invocation. Many commands invoked '\
+                  'with this option will just crash.'
```
danielbreves: > Many commands invoked with this option will just crash.

Is this because the command expects input that was not passed by the original invocation? I don't find that statement very helpful 😅

sandlerr: yes, but we also might not have a mechanism to pass all the input to all the commands

lib/zendesk_apps_tools/command_helpers.rb
sandlerr: yes

lib/zendesk_apps_tools/common.rb
```diff
@@ -20,17 +55,18 @@ def get_value_from_stdin(prompt, opts = {})
...
-        end
+        break if input.to_s =~ options[:valid_regex]
+        say_error options[:error_msg]
       end
 
       input
     end
 
     def get_password_from_stdin(prompt)
+      if options[:unattended]
```
danielbreves: Don't need to check `respond_to?(:options)` here?

lib/zendesk_apps_tools/common.rb
```diff
@@ -10,6 +41,10 @@ def say_error(msg)
     end
 
     def get_value_from_stdin(prompt, opts = {})
+      if respond_to?(:options) && options[:unattended]
+        say_error 'Would have prompted for a value interactively, but we are running unattended.'
+        say_error_and_exit prompt
```
sandlerr: I disgaree. We're not ignoring, we're crashing. Also, that's going to output some really long lines.

lib/zendesk_apps_tools/common.rb
sandlerr: good catch, this is included in a module that won't have options. Needs a failing test too.

lib/zendesk_apps_tools/common.rb
danielbreves: Ok

lib/zendesk_apps_tools/command_helpers.rb
sandlerr: sorry, I misunderstood here. `Command` includes `CommandHelpers`, not directly `Common` - this needs to be in both places so that classes that just include `Common` also get the ClassMethods.

lib/zendesk_apps_tools/common.rb
danielbreves: `--not-interactive`?

lib/zendesk_apps_tools/common.rb
danielbreves: How about "This option may not work with some commands", and bonus points for adding the list of commands that don't work? 😸

sandlerr: if we were going to go that far we'd not even make the option available on known broken commands. I was going to avoid that work, while allowing people to benefit if it does happen to work in some cases.

lib/zendesk_apps_tools/common.rb
sandlerr: That would be ok, mind if we go with a positive option rather than a negative option?

lib/zendesk_apps_tools/common.rb
sandlerr: Added "Experimental"

PR #152, opened 2016-10-21
Adding support for binding using zat server

Typhon66:
Adding support to allow binding a specific ip address when using zat server. This would be useful for our case when developing using a vagrant/virtual machine, as we tend to bind our tools to 0.0.0.0 when testing and setup our virtual machine to allow access to the virtual machine through a specific ip/host.


Typhon66:
also, set it up to default to 127.0.0.1 so if people don't use this functionality, they can continue to just run `zat server` without binding. But if you did want to bind it, you could run `zat server --bind=0.0.0.0`

sandlerr:
This is released in [v1.35.6](https://github.com/zendesk/zendesk_apps_tools/releases/tag/v1.35.6).

PR #158, opened 2016-11-12
Revert "name is not a thing"

ocke:
:v:

/cc @zendesk/vegemite

### Description
This reverts commit c51079ff7cf84fa0b04ab9b0a113308079d39702.

Make name  a thing again!

https://github.com/zendesk/zendesk_apps_tools/pull/137#issuecomment-260105744

We have to merge https://github.com/zendesk/zendesk_apps_support/pull/141 first before this works.

### References
* JIRA: 

### Risks

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ PATH
       sinatra (~> 1.4.6)
       sinatra-cross_origin (~> 0.3.1)
       thor (~> 0.19.1)
-      zendesk_apps_support (~> 3.1.5)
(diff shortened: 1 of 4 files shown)


sandlerr:
won't work this way, need to use package.original_manifest I think

ocke:
I see there is a manifest class now. Which is def 👍 

I added name to the class: https://github.com/zendesk/zendesk_apps_support/pull/141

sandlerr:
ok, in that case this pr is blocked behind bumping and releasing zas

PR #159, opened 2016-11-14
bump zendesk apps support to 3.1.5

token-cjg:
@zendesk/vegemite 

Bumps the zendesk apps support gem from 3.1.4 to 3.1.5 to support the new ticket_editor location.

**Risks**:

Might introduce functionality not yet properly supported in production (the new ticket_editor location).

**Mitigations**:

This should be okay to go, pending a green light on Jenkins.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ PATH
       sinatra (~> 1.4.6)
       sinatra-cross_origin (~> 0.3.1)
       thor (~> 0.19.1)
-      zendesk_apps_support (~> 3.1.4)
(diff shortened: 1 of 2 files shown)


sandlerr:
This needs to be updated: https://github.com/zendesk/zendesk_apps_tools/blob/master/zendesk_apps_tools.gemspec#L22

Gemfile.lock is only used during development and tests, not for installed copies of the gem.

token-cjg:
Thanks for the pointer, I think I've bumped the right part of the project now.  Feel free to let me know if there is anything else I've missed.

PR #161, opened 2016-11-16
Add new default assets

danielbreves:
:v:

Deletes banner.png and logo-promotion.png and replaces logo-small.png and logo.png.

logo-small
logo

/cc @zendesk/vegemite


sandlerr:
what's up with the size change?

danielbreves:
@sandlerr I think the old one was wrong, this is what we document [here](https://developer.zendesk.com/apps/docs/agent/styling#assets).

PR #162, opened 2016-11-17
Fix broken link

danielbreves:
:v:

/cc @zendesk/vegemite


adammw:
Technically not that broken... having the whole section documentation also means it's less likely for the link to break (section renames automatically handle redirects to the new url)

danielbreves:
Ah right, I sort of assumed this was broken like the others. Closing.

PR #163, opened 2016-11-28
support --unattended in zat server, read default parameters instead of crashing

sandlerr:
I think it's better to do this because foreman in the app scaffold will dump a bunch of webpack output after asking for a setting, this caused confusion.

cc @zendesk/vegemite


adammw:
Probably needs a bit more documentation, took me a while to understand the change here...

PR #166, opened 2016-12-07
Fix write directory for translations

fvzzy:
/cc @zendesk/undefined

### Description
When specifying an optional path flag, prevent the root directory name being recreated and nested i.e. stop stuff like `/dist/dist/translations/` happening

### References
* JIRA: AF-564

### Risks


mariasica:
@zendesk/i18n-dev

spec/lib/zendesk_apps_tools/translate_spec.rb
```diff
@@ -89,6 +89,7 @@
       allow(translate).to receive(:create_file)
 
       expect(translate).to receive(:nest_translations_hash).once.and_return({})
+      expect(translate).to receive(:write_json).once.with("translations/en.json", {})
```
sandlerr: never mind, the file doesn't exist, its stubs all the way down

PR #167, opened 2016-12-14
Add default bar app icons

adammw:
:v:

/cc @zendesk/vegemite

### Description

Add default bar app icons

### Risks
* low: too many people submit apps with default icons when they don't need them


sandlerr:
why? lotus will serve defaults if they're missing

adammw:
Doesn't seem to happen in ZAT, plus it gives devs a base to work off?

sandlerr:
at the cost of making it impossible to change the default later

adammw:
Heh heh... yeah.

PR #170, opened 2016-12-23
dont crash when the mtime is nil for some reason

sandlerr:
cc @zendesk/vegemite


sandlerr:
avoids `comparison of Time with nil failed`

PR #173, opened 2017-01-11
Make app_id default to a numeric value instead of a string

swmichaels:
@zendesk/vegemite 

Hey guys, if we have a line in an app like the [following](https://github.com/zendesk/timetracking_app/blob/master/app.js#L66):

```onAppCreated: function() {
  if (this.installationId()) {
    ... blah blah here
  }
```

if you are running the app from zat (which I think this conditional is checking for), the installationId() method was returning a string '0' by default, which is truthy. I've changed ZAT to only use a default app_id of the numeric 0 where necessary. I suppose I could just do `parseInt(this.installationId(), 10)` instead, but I wanted to see why this was happening at the source.

I took a look at Thor's [Method Options](https://github.com/erikhuda/thor/wiki/Method-Options) docs as a reference for forcing a numeric type.

Thoughts?

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -121,13 +121,13 @@ module ZendeskAppsTools
     DEFAULT_SERVER_PATH = './'
     DEFAULT_CONFIG_PATH = './settings.yml'
     DEFAULT_SERVER_PORT = '4567'
-    DEFAULT_APP_ID = '0'
+    DEFAULT_APP_ID = 0
 
     desc 'server', 'Run a http server to serve the local app'
     shared_options(except: [:clean])
     method_option :config, default: DEFAULT_CONFIG_PATH, required: false, aliases: '-c'
     method_option :port, default: DEFAULT_SERVER_PORT, required: false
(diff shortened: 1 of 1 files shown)


ocke:
The app should really use `this.isZatEnabled()` and don't rely on the fact that it is ID 0, which happens to be only true for Zat. Sam for `appId()`

sandlerr:
@ocke yes it should and I have a PR open to do that here: https://github.com/zendesk/timetracking_app/pull/78 but zat should still be consistent here with normal apps.

swmichaels:
Isn't it possible to assign app_id to any arbitrary object? I haven't tested- just interested. The possibility of a string showing up was pretty confusing to me when trying to debug the time tracking app 💣

sandlerr:
yep, you can currently type whatever you like in the commandline for `app_id`. Even with this patch it will accept decimals, but you'd have to go out of your way to shoot yourself in the foot that way.

sandlerr:
Actually, we shouldn't use `isZatEnabled` because you might want to test in ZAT against a real app's requirements.

sandlerr:
just released this in 1.37.4 sorry for the delay

PR #174, opened 2017-01-12
add a bit more installation instructions now that we depend on a C ex…

sandlerr:
…tension via sassc

cc @zendesk/vegemite


sandlerr:
this has been added to docs following the links from https://developer.zendesk.com/apps/docs/apps-v2/getting_started#zendesk-app-tools by @chucknado

PR #176, opened 2017-02-28
Set install_id from ZAT

ocke:
:v:

/cc @zendesk/vegemite

### Description
You can set the app id, but not the install id. This PR allows you to set the install id. Useful for using the proxy and testing currently installed apps locally.

### References
* JIRA: 

### Risks
- [low] technically when using app_id it doesn't set install_id anymore.

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -153,7 +153,7 @@ module ZendeskAppsTools
         server.set :public_folder, File.join(options[:path], 'assets')
         server.set :parameters, settings
         server.set :app_id, options[:app_id]
-        server.set :install_id, options[:install_id]
+        server.set :install_id, options[:install_id] || options[:app_id] # backwards compatible
         server.run!
       end
     end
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_tools/server.rb
```diff
@@ -17,7 +17,7 @@ class Server < Sinatra::Base
       package = ZendeskAppsSupport::Package.new(settings.root, false)
       app_name = package.manifest.name || 'Local App'
       installation = ZendeskAppsSupport::Installation.new(
-        id: settings.app_id,
+        id: settings.install_id,
```
sandlerr: how about `settings.install_id || settings.app_id` for people already using it?

ocke: I fixed it in command.rb

lib/zendesk_apps_tools/command.rb
```diff
@@ -151,6 +153,7 @@ def server
         server.set :public_folder, File.join(options[:path], 'assets')
         server.set :parameters, settings
         server.set :app_id, options[:app_id]
+        server.set :install_id, options[:install_id] != 0 ? options[:install_id] : options[:app_id] # backwards compatible
```
sandlerr: nitpick: replace `0` by `DEFAULT_INSTALL_ID`

PR #180, opened 2017-03-09
create_file regression test

ocke:
:v:

/cc @zendesk/vegemite

### Description
Added a regression test for translate update

### References
* JIRA:

### DIFF
--- a/spec/lib/zendesk_apps_tools/translate_spec.rb
+++ b/spec/lib/zendesk_apps_tools/translate_spec.rb
@@ -103,34 +103,36 @@ describe ZendeskAppsTools::Translate do
 
   describe '#write_json' do
     let(:translate) { ZendeskAppsTools::Translate.new }
-    let(:test_file_path) { '/tmp/test.json' }
 
     before do
       allow(translate).to receive(:options) { { unattended: true } }
     end
 
-    after do
-      File.delete(test_file_path)
+    around(:example) do |example|
+      Dir.mktmpdir do |dir|
+        @test_file_path = "#{dir}/test.json"
+        example.run
+      end
     end
 
     it 'works for identical ascii' do
-      translate.create_file(test_file_path, JSON.pretty_generate({ node: "test abc" }) + "\n", force: true)
-      expect { translate.write_json(test_file_path, { node: "test abc" }) }.to output("   identical  #{test_file_path}\n").to_stdout
+      translate.create_file(@test_file_path, JSON.pretty_generate({ node: "test abc" }) + "\n", force: true)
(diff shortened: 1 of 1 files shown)


spec/lib/zendesk_apps_tools/translate_spec.rb
```diff
@@ -101,6 +101,39 @@
...
 
+  describe '#write_json' do
+    let(:translate) { ZendeskAppsTools::Translate.new }
+    let(:test_file_path) { '/tmp/test.json' }
+
+    before do
+      allow(translate).to receive(:options) { { unattended: true } }
+    end
+
+    after do
```
sandlerr: how about

```ruby
around(:example) do |example|
  Dir.mktmpdir do |dir|
    @test_file_path = "#{dir}/test.json"
    example.run
  end
end
```

sandlerr: that way, if for some reason the test fails to create the file we don't also get errors in the after block, and we don't have to write an existence check.

spec/lib/zendesk_apps_tools/translate_spec.rb
```diff
@@ -101,6 +101,41 @@
...
+    around(:example) do |example|
+      Dir.mktmpdir do |dir|
+        @test_file_path = "#{dir}/test.json"
+        example.run
+      end
+    end
+
+    it 'works for identical ascii' do
+      translate.create_file(@test_file_path, JSON.pretty_generate({ node: "test abc" }) + "\n", force: true)
+      expect { translate.write_json(@test_file_path, { node: "test abc" }) }.to output("   identical  #{@test_file_path}\n").to_stdout
```
sandlerr: it'll probably be a little less flaky if we just call [`lstrip`](https://ruby-doc.org/core-2.4.0/String.html#method-i-lstrip) rather than relying on it outputting exactly this many white spaces?

ocke: sure, but how about the spaces between the word `identical`/`force` and the file name?

sandlerr: I wouldn't mind leaving the filename in there in case we mess that up somehow. We could `split` the string on whitespace? 🤷‍♂️

ocke: actually, it's always 2 spaces. so i'll just lstrip

ocke: actually, how can we lstrip?

sandlerr: hm, we'd have to re-write this without the rpsec nicety to capture stdout as a StringIO then read it as a String to lstrip: http://stackoverflow.com/a/22777806/1911487

Since it adds extra code I'm happy not to bother

PR #182, opened 2017-03-23
just call validate rather than invoking

sandlerr:
### Description

`zat create --no-install` crashes because it passes the `--no-install` argument on to `validate`.

cc @zendesk/vegemite 

### References

https://zendesk-appdevelopers.slack.com/archives/C0L7VFW9G/p1490251653651834

### DIFF
--- a/features/step_definitions/app_steps.rb
+++ b/features/step_definitions/app_steps.rb
@@ -57,7 +57,6 @@ end
 
 When /^I run the command "(.*?)" to (validate|package|clean|create) the app$/ do |cmd, _action|
   env_hash = prepare_env_hash_for(cmd)
-  cmd.sub!(/^zat/, File.expand_path('../../../bin/zat', __FILE__)) unless ENV['TRAVIS']
   IO.popen(env_hash, cmd, 'w+') do |pipe|
     pipe.puts "\n"
     pipe.close_write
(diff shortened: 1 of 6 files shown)


features/step_definitions/app_steps.rb
```diff
@@ -49,8 +55,15 @@
   FileUtils.ln_s(src, dest)
 end
 
-When /^I run the command "(.*?)" to (validate|package|clean) the app$/ do |cmd, _action|
-  IO.popen(cmd, 'w+') do |pipe|
+When /^I run the command "(.*?)" to (validate|package|clean|create) the app$/ do |cmd, _action|
+  if cmd.start_with? 'zat create'
```
danielbreves: Can extract this to a named method? `stub_api_requests`?

features/support/helpers.rb
```diff
@@ -0,0 +1,8 @@
+def prepare_env_hash_for(cmd)
+  if cmd.start_with? 'zat create'
+    path = File.expand_path('../../support/webmock', __FILE__)
+    { 'RUBYOPT' => "-r #{path}" }
```
danielbreves: This seems to be a bit more specific than what the method name implies, no?

features/support/webmock.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+unless defined?(Cucumber)
```
danielbreves: Worth a comment explaining why this is here?

features/support/helpers.rb
sandlerr: ? it returns a hash of environment variables :/

danielbreves: It does, but only one specific env var for one specific command.

sandlerr: yes, we lost this information when we chose to extract it.

`prepare_env_hash_for_stubbing_requests_in_zat_create` ?

PR #185, opened 2017-04-11
Travis: Add Sourceclear security scan to CI

imohammedzen:
### Context
Source Clear helps us to find vulnerabilities in third party components ( open source libraries) which we consume in our code base. Once we identify issues, we can manage them using Source Clear SaaS portal.


imohammedzen:
@svizzari @sandlerr could you please add SRCCLR_API_TOKEN under settings at https://travis-ci.org/zendesk/zendesk_apps_tools/. Thanks.

PR #187, opened 2017-05-16
theming!

sandlerr:
Provides local theming experience for Theming Center customers (Beta).

It:
- loads theme files from local filesystem
- pushes some of the files to an internal endpoint
- serves some others locally
- detects file changes: triggers new upload and page reload

@zendesk/vegemite @zendesk/delta


spec/spec_helper.rb
```diff
@@ -6,6 +8,7 @@
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 
 require 'webmock/rspec'
+require 'zendesk_apps_tools'
```
sandlerr: the actual app does this to setup autoload paths, so we should do it in tests. Otherwise I'm ending up in a circular require trap.

lib/zendesk_apps_tools/theme.rb
```diff
@@ -40,8 +40,8 @@ def generate_payload
           asset_payload[File.basename(asset)] = url_for(asset)
         end
         payload['assets'] = asset_payload unless asset_payload.empty?
-        payload['js'] = javascript if javascript
-        payload['css'] = stylesheet_content if stylesheet_content
+        payload['js'] = url_for(theme_package_path('script.js')) if File.file?(theme_package_path('script.js'))
```
svizzari: Good for completeness, but [doesn't look like this is being used](https://github.com/zendesk/help_center/pull/13668/files#diff-127a850b7d7b20173ad3fba88f8a25e3R134) on the HC side.  Same for CSS

lib/zendesk_apps_tools/command.rb
```diff
@@ -10,13 +10,16 @@ class Command < Thor
...
 
-    map %w(-v) => :version
+    map %w[-v] => :version
 
     source_root File.expand_path(File.join(File.dirname(__FILE__), '../..'))
 
     desc 'translate SUBCOMMAND', 'Manage translation files', hide: true
     subcommand 'translate', Translate
 
+    desc 'theme SUBCOMMAND', 'Development tools for Zendesk Themes', hide: true
```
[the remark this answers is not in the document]
augustocravosilva: We can make it visible but we should add disclaimer that this is beta
I'll add a commit with it

ocke:
rake wouldn't work until i changed to sinatra 2.0.0 :man_shrugging:

libo:
@augustocravosilva could you please give it a try and see it this is working as expected.

augustocravosilva:
@ocke Live reload is now broken.
After the first upload, I get this error:
``` ruby
NotImplementedError - only partial hijack is supported.:
```

This seems to be because you removed `thin`. `webbrick` is not supported by the websocket gem that is being used. See https://github.com/faye/faye-websocket-ruby#running-your-socket-application

augustocravosilva:
@zendesk/vegemite Maybe a question for all:
> I added a commit that switches to thin at a global level. Is that ok with you guys?

danielbreves:
Apologies for the delay. I don't want to block this, but were you guys planning to add some tests?

danielbreves:
I tested `zat new/server/validate/package/create/update`. Seems to work fine 👍

sandlerr:
🎉 :shipit:

PR #189, opened 2017-06-13
at least try to serve the app even if assets/app.js exists

sandlerr:
### Description

Sinatra's static file serving seems to take precedence over our defined routes. This is bad in the case that we want to _generate_ `app.js` rather than just serve it from `assets/`.

Helpfully, we've already monkey patched its `send_file` method, so let's monkey patch some more.

cc @zendesk/vegemite @dashedstripes 

### Risks

- [low] `zat server` is broken for some apps


ocke:
you mean that `/assets/app.js` will take preference over the generated `/app.js`?

sandlerr:
> you mean that /assets/app.js will take preference over the generated /app.js?

it does right now in master. I'm trying to serve our generated app.js when the query parameters are there.

PR #190, opened 2017-06-19
[AF-682] Add sunset warning when running zat update

nick-f:
:v:

/cc @zendesk/vegemite

### Description
Displays a warning message if the manifest framework version is 1.0.

~This will also require a PR to ZAS to [update the version numbers here](https://github.com/zendesk/zendesk_apps_support/blob/bb88414ed71704a69ffdcfed2736179600ad4a26/lib/zendesk_apps_support/app_version.rb#L12-L14).~ ZAS PR: https://github.com/zendesk/zendesk_apps_support/pull/166

zat_update

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-682
* ZAS PR: https://github.com/zendesk/zendesk_apps_support/pull/166

### Risks
* [low] Message being added to terminal output

PR #192, opened 2017-06-21
[AF-746] Fix default translation files in ZAT

ocke:
:v:

/cc @zendesk/vegemite

### Description
Fix V2 and V1 default translation files.

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-746

### Risks
[low] play button still doesn't work

### DIFF
--- a/app_template/translations/en.json
+++ b/app_template/translations/en.json
@@ -1,7 +1,8 @@
 {
   "app": {
-    "description": "Play the famous zen tunes in your help desk.",
-    "name": "Buddha Machine"
+    "short_description": "Play the famous zen tunes in your help desk.",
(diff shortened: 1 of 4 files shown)


svizzari:
What do you think about the idea of somehow using templates and injecting the short_description, installation_instructions, etc. in, so that they're only stored in one place?

ocke:
you mean for v1 and v2, or all the code in general the fact we have it 4 times now.

ocke:
We will get rid of V1 soon anyway, so then that code will be removed.

nick-f:
@svizzari The app_template folder (the v1 template) is going to be removed with [the deprecation warning PR](https://github.com/zendesk/zendesk_apps_tools/pull/190) if there's no objections or other changes required.

PR #195, opened 2017-07-07
Remove command for new v1 app from docs and add sunsetting message

nick-f:
:v:

/cc @zendesk/vegemite

### Description
Update docs to remove references to removed feature and show sunsetting message instead.

### Risks
* [none] Docs update.


nick-f:
Updating the method option description updates the help text.

image

PR #200, opened 2017-08-14
Add version checking

nick-f:
/cc @zendesk/vegemite 

## Description
Add version checking to ZAT.

When a check is made, the version number and the date it was last checked are both saved to the `.zat` file for caching and so it wouldn't happen every time the validate command was called. Currently it will only check if the last time a check was made was more than a week ago.

Checks are performed by querying the RubyGems API for the version.

The cached values are only saved to the project's local `.zat` file, as opposed to a global, root `.zat` file. Maybe that could be a future feature improvement.

## Steps to test
- Check out `nickf/add-version-checks` branch
- `bundle i`
- If you haven't already, set up an alias in your shell profile (.bash_profile, .zshrc, whatever you use), and change the path in the command  
for example: `alias zatdev='BUNDLE_GEMFILE=/path/to/repo/zendesk_apps_tools/Gemfile bundle exec /path/to/repo/zendesk_apps_tools/bin/zat'`  
You might need to `source` your profile to refresh the changes (or open a new Terminal window)
- Go to a folder that contains an app
- `zatdev validate`
- See that the check is performed and results are appended to `.zat`
- `zatdev validate` again
- See that the checks aren't performed again

### DIFF
--- a/features/create.feature
+++ b/features/create.feature
@@ -5,6 +5,7 @@ Feature: upload an app to the apps marketplace
     Given an app is created in directory "tmp/aruba"
     Given a .zat file in "tmp/aruba"
     When I run the command "zat create --path tmp/aruba --no-install" to create the app
+    And the command output should contain "info  Checking for new version of zendesk_apps_tools"
     And the command output should contain "validate  OK"
     And the command output should contain "package  created"
     And the command output should contain "Status  working"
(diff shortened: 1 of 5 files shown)


danielbreves:
Nice!! 👍  on 💚

fvzzy:
Sweet, this is cool. Does this need another test in `https://github.com/zendesk/zendesk_apps_tools/blob/master/spec/lib/zendesk_apps_tools/common_spec.rb` though? Also, is `common.rb` the right place for it? Seems like mostly higher level utils
 in there. Otherwise 👍 .

nick-f:
@fvzzy I couldn't think of a better place to put it. It's called when validate is run (which is also when package is run). Part of the reasoning was we might want to put it in other places later? Happy to move it though, just not sure where.

fvzzy:
@nick-f it's own module maybe? Or as a protected method in `command.rb`, where things like `setup_path` live? 

A nitpick in any event and not worth holding up the PR for. I do think whichever module it lives in could be extended with a test for it though.

nick-f:
@fvzzy 💇‍♂️ Moved to protected method in `command.rb` and added tests.

alieu:
:+1:

PR #201, opened 2017-08-22
Warn user about manifest.json parsing errors

stevenpetryk:
Dear beloved Zendesk,

I was migrating my application from v1 to v2, and goofed when I was editing my `manifest.json`. I had accidentally duplicated the `locations` key. Oops!

The good news is that there's already custom parser logic in `zendesk_apps_support` for handling duplicated keys. The bad news is that this gem completely ignores those exceptions and instead always says `Manifest file cannot be found in the given path. Check you are pointing to the path that contains your manifest.json`. But no longer, for it is fixed!

**In the case of duplicated keys:**

```
Manifest file is incorrectly formatted: Duplicate reference in manifest: "locations". Initially set to "[1]", attempted overwrite to "[2]".
```

**In the case of invalid formatting**

```
Manifest file is incorrectly formatted: 757: unexpected token at '{ "locations": [1] "locations": [2] }'
```


stevenpetryk:
just casually bumping this after 2 years

ocke:
I couldn't fix the conflict because the branch was deleted: https://github.com/zendesk/zendesk_apps_tools/pull/284

PR #203, opened 2017-09-05
Add require statement for ExecJS

nick-f:
/cc @zendesk/vegemite 

### Description
Stops `zat validate` from completely falling over.

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -63,6 +63,7 @@ module ZendeskAppsTools
     desc 'validate', 'Validate your app'
     shared_options(except: [:unattended])
     def validate
+      require 'execjs'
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_tools/command.rb
```diff
@@ -63,6 +63,7 @@ def new
     desc 'validate', 'Validate your app'
     shared_options(except: [:unattended])
     def validate
+      require 'execjs'
```
hoangtrvu: other than requiring it here, I think we should include it as a dependency here as well https://github.com/zendesk/zendesk_apps_tools/blob/master/zendesk_apps_tools.gemspec#L23

nick-f: 💇‍♂️

PR #205, opened 2017-09-08
Handle exceptions for duplicate keys in manifest.json during zat create and validate

pchhetri:
When running `zat create` or `zat validate` exception is thrown when the manifest contains duplicate keys.

```
{
  "name": "foobar",
  "name": "foobar",
  ...
}
```
```
▶ bundle exec ruby bin/zat validate --path=../foobar
    validate  Duplicate reference in manifest: "name". Initially set to "foobar", attempted overwrite to "foobar".

▶ bundle exec ruby bin/zat create --path=../foobar
       error  Duplicate reference in manifest: "name". Initially set to "foobar", attempted overwrite to "foobar".
```

This hopes to address ZD [#2895995](https://support.zendesk.com/agent/tickets/2895995)

@zendesk/vegemite


ocke:
not a bad idea, so we still fail the validate, but it doesn't throw an ruby exception anymore.

Can you make the tests green? :D

ocke:
@pchhetri

pchhetri:
cc: @zendesk/vegemite this should be ready to go now. Although, there is a case where it might fail, this concern is raised in [ZD 2895995](https://support.zendesk.com/agent/tickets/2895995).

features/create.feature
```diff
@@ -11,3 +11,19 @@ Feature: upload an app to the apps marketplace
...
     And the command output should contain "Create  OK"
     And the zip file should exist in directory "tmp/aruba/tmp"
+
+  Scenario: invalid app - missing manifest.json
+    Given an app is created in directory "tmp/aruba"
+    Given I remove file "tmp/aruba/manifest.json"
+    When I run the command "zat create --path tmp/aruba --no-install" to validate the app
+    Then the command output should contain "Could not find manifest.json"
+
+  Scenario: invalid manifest.json - duplicate keys
```
danielbreves: The tests below seem redundant, why not only keep the `validate` ones?

pchhetri: iirc zat create accesses the [manifest](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/command.rb#L164-L171) before running validate (via [package](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/command.rb#L164-L171)). I added these tests so that if there are problems with manifest, it would cover those scenarios.

danielbreves:
@pchhetri we may need to `autoload` `ZendeskAppsSupport::Manifest::OverrideError` [here](https://github.com/zendesk/zendesk_apps_support/blob/8a3101181c8c7023314f4e95659dcc35c6886017/lib/zendesk_apps_support.rb#L33-L37).

Lobaster:
this appears to be handled by this merged PR:
https://github.com/zendesk/zendesk_apps_support/pull/190

hoangtrvu:
Closing since this has been fixed by another PR [here](https://github.com/zendesk/zendesk_apps_support/pull/190).

PR #207, opened 2017-09-15
[AF-775] Create ZAT migrate command

svizzari:
✌️ 

/cc @zendesk/vegemite 

### Overview
Adds support for a `zat migrate` command that will shell out to the [Zendesk App Migration Helper](https://github.com/zendesk/zendesk_app_migrator).

Before attempting to run the Migration Helper, zat will:
1. Check whether Node.js is installed (and by extension, NPM)
1. Check whether the App Migration Helper is installed
1. Run the App Migration Helper (with any options passed to zat)

### Todo
- [x] Add tests
- [x] Add support for `--replace-v1` option cf5adf2

### References
- https://zendesk.atlassian.net/browse/AF-775

### Risks
- N/A

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -193,6 +193,7 @@ module ZendeskAppsTools
 
     desc 'migrate', 'Helps with the migration of a v1 app to v2'
     shared_options(except: [:unattended, :clean])
+    method_option %s(replace-v1), default: false, required: false, type: :boolean, aliases: '-r'
     method_option :auto, default: false, required: false, type: :boolean, aliases: '-a'
     def migrate
       cache.clear
@@ -248,6 +249,7 @@ module ZendeskAppsTools
 
     def migrate_app(options = {})
       cmds = ["app_migrator", "migrate", "--path=#{options[:path]}"]
+      cmds << "--replace-v1" if options[:"replace-v1"]
       cmds << "--auto" if options[:auto]
       system(*cmds)
     end
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -200,7 +200,7 @@ module ZendeskAppsTools
       @command = 'Migrate'
       unless migration_helper_installed
         try_install = get_value_from_stdin("The Zendesk App Migration Helper isn't installed yet. Would you like to try installing now?", limited_to: ['y', 'yes', 'n', 'no'], default: 'y' )
-        say_error_and_exit("Please install the Zendesk App Migration Helper before running this command") unless try_install =~ /^y(es)?$/i
+        say_error_and_exit("Please install the Zendesk App Migration Helper before running this command") if (try_install =~ /^y(es)?$/).nil? # thats a no
         say_error_and_exit("Unable to install the Zendesk App Migration Helper \
                             Please follow the installation instructions at \
                             https://github.com/zendesk/zendesk_app_migrator \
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_tools/command.rb
```diff
@@ -191,13 +191,67 @@ def update
...
     end
 
     protected
 
+    def migration_helper_installed
+      is_semver(app_migrator_version)
+    end
+
+    def node_js_installed
+      !!system("npm -v", out: File::NULL)
```
hoangtrvu: do we have any requirements for node/npm version?

lib/zendesk_apps_tools/command.rb
```diff
@@ -191,13 +191,67 @@ def update
...
+      !(/^v?(\d+\.)?(\d+\.)?(\*|\d+)$/ =~ version).nil?
+    end
+
+    def install_migration_helper
+      system("npm install -g zendesk_app_migrator", out: File::NULL)
+    end
+
+    def migrate_app(options = {})
+      cmds = ["app_migrator", "migrate", "--path=#{options[:path]}"]
+      cmds << "--auto" if options[:auto]
```
hoangtrvu: should we passthrough `-r --replace-v1` flag as well?

lib/zendesk_apps_tools/command.rb
svizzari: Good question.  I'd incorrectly assumed that the `.node-version` config would be respected when the package was globally installed.  This not being the case, I feel like using the [`engines` field in package.json](https://docs.npmjs.com/files/package.json#engines) might be the best approach.  Would you agree?

lib/zendesk_apps_tools/command.rb
svizzari: 💇 cf5adf2

lib/zendesk_apps_tools/command.rb
```diff
@@ -191,13 +191,69 @@ def update
...
+    method_option %s(replace-v1), default: false, required: false, type: :boolean, aliases: '-r'
+    method_option :auto, default: false, required: false, type: :boolean, aliases: '-a'
+    def migrate
+      cache.clear
+      setup_path(options[:path])
+      @command = 'Migrate'
+      unless migration_helper_installed
+        try_install = get_value_from_stdin("The Zendesk App Migration Helper isn't installed yet. Would you like to try installing now?", limited_to: ['y', 'yes', 'n', 'no'], default: 'y' )
+        say_error_and_exit("Please install the Zendesk App Migration Helper before running this command") if (try_install =~ /^y(es)?$/).nil? # thats a no
+        say_error_and_exit("Unable to install the Zendesk App Migration Helper \
```
danielbreves: nitpick: I think it would read nicer if it was `install_migration_helper` then checking `migration_helper_installed` below.

lib/zendesk_apps_tools/command.rb
```diff
@@ -191,13 +191,69 @@ def update
...
+
+    def node_js_installed
+      !!system("npm -v", out: File::NULL)
+    end
+
+    def app_migrator_version
+      say_error_and_exit("Node.js and NPM are required to use the Zendesk App Migration Helper \
+                          Please see installation instructions at https://nodejs.org/en/download/") unless node_js_installed
+      @migrator_version ||= begin
+        IO.popen("npm ls -g --depth=0 | grep zendesk_app_migrator") do |ls_io|
```
danielbreves: Could we make this easier by exposing a `-v` command in the migrator?

svizzari: Actually, the migrator already has a `--version` option, and there is [a PR to add `-v`](https://github.com/zendesk/zendesk_app_migrator/pull/31).

As discussed in 🍖 -space, I'll refactor slightly in this PR to use `system` to determine whether the migrator is installed, rather than `IO.popen`

svizzari:
@danielbreves 💇 Comments addressed

spec/lib/zendesk_apps_tools/command_spec.rb
```diff
@@ -192,4 +192,63 @@
...
+      allow(@command).to receive(:get_value_from_stdin)
+      allow(@command).to receive(:install_migration_helper)
+      allow(@command).to receive(:say_error_and_exit)
+      allow(@command).to receive(:migrate_app)
+    end
+
+    context 'when node and npm are not installed' do
+      before do
+        allow(@command).to receive(:package_installed).with('npm').and_return(false)
+        allow(@command).to receive(:package_installed).with('app_migrator').and_return(false)
```
hoangtrvu: /nitpick this line is not necessary

hoangtrvu:
@svizzari I added 1 commit https://github.com/zendesk/zendesk_apps_tools/pull/207/commits/25344d45a10ec32e9d5c077726924b36352615ca to update the tests, let me know if you are okay with it

svizzari:
@hoangtrvu 

> I added 1 commit 25344d4 to update the tests to cover more cases, let me know if you are okay with it

Looks good to me, I had a couple of minor stylistic comments only.  I think we're good to merge here

PR #208, opened 2017-09-22
Prevent multiple installations in one product

fbvilela:
🐻 

@zendesk/vegemite @zendesk/wombat 

### Description

Phil found a bug where a `zat create` would print ` installing in support` twice.
it was because the app had two locations in support. 

Peeps, I think the tests on zat have room for improvement. It seems to be stubbing out too many things. It's hard to figure out what's really testing. It might be worth creating a story as Tech Debt to refactor some of the tests ( at least the command_spec ) to use real `ZendeskAppsSupport` objects like I just did in this PR

### Risks
* low - breaks `zat create` app installation


spec/fixture/two_locations_manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+    "author": {
+      "name": "Zendesk",
+      "email": "support@zendesk.com"
+    },
+    "location": [
```
ocke: you're testing a correct manifest? Isn't this manifest supposed to be broken?

fbvilela: it's happening to one of our official apps. This manifest is a copy/paste.

hoangtrvu:
Is this PR still valid @fbvilela? If yes, can you rebase it 🙇?

PR #218, opened 2017-11-22
Add option to provide a list of locales to fetch

baopham:
We currently fetch the locales from https://support.zendesk.com/api/v2/locales.json. However, other products may have more languages than Support.

Following the approach taken in https://github.com/zendesk/client-i18n-tools, adding a new option `--locales=./path/to/locales.json` to `translate update` command for the user to provide a list of locales.

> User can pass in absolute or relative path of the json file

Existing usage of `zat translate update` should still be working. Current logic is untouched.

@zendesk/vegemite - please help release a new version too as it will help unblock a few things. Many thanks!

### DIFF
--- a/lib/zendesk_apps_tools/translate.rb
+++ b/lib/zendesk_apps_tools/translate.rb
@@ -10,7 +10,8 @@ module ZendeskAppsTools
     include Thor::Actions
     include ZendeskAppsTools::Common
 
-    LOCALE_ENDPOINT = 'https://support.zendesk.com/api/v2/locales/agent.json'
+    LOCALE_BASE_ENDPOINT = 'https://support.zendesk.com/api/v2/locales'
+    LOCALE_ENDPOINT = "#{LOCALE_BASE_ENDPOINT}/agent.json"
 
     desc 'to_yml', 'Create Zendesk translation file from en.json'
     shared_options(except: [:clean])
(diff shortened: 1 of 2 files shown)


lib/zendesk_apps_tools/translate.rb
```diff
@@ -10,7 +10,8 @@ class Translate < Thor
     include Thor::Actions
     include ZendeskAppsTools::Common
 
-    LOCALE_ENDPOINT = 'https://support.zendesk.com/api/v2/locales/agent.json'
+    LOCALE_BASE_ENDPOINT = 'https://support.zendesk.com/api/v2/locales'
+    LOCALE_ENDPOINT = LOCALE_BASE_ENDPOINT + '/agent.json'
```
ocke: `"#{LOCALE_BASE_ENDPOINT}/agent.json"`

lib/zendesk_apps_tools/translate.rb
```diff
@@ -173,6 +175,15 @@ def locale_list
         say('Fetching translations...')
         require 'net/http'
         require 'faraday'
+
+        if options[:locales]
+          content = File.read(File.expand_path(options[:locales]))
+          locales = JSON.parse(content)
+          return locales.map do |locale|
+            { 'locale' => locale, 'url' => LOCALE_BASE_ENDPOINT + "/#{locale}.json" }
```
ocke: `{ 'locale' => locale, 'url' => "#{LOCALE_BASE_ENDPOINT}/#{locale}.json" }`

Issue #223, opened 2017-12-22
error on zat package


pedrorenan:
Hi, when I try use zat package I receive this message

/Library/Ruby/Gems/2.3.0/gems/zendesk_apps_tools-2.8.1/lib/zendesk_apps_tools/command.rb:280:in `check_for_update': uninitialized constant ZendeskAppsTools::Command::DateTime (NameError)

jbescoyez:
I have the same issue. @pedrorenan Have you found a workaround?

PolRod:
Exact same error here for `zat validate`.

pedrorenan:
Hi! 

I edited the file commenting the line that accused the error. I managed to use it quietly later.

PolRod:
I added `require 'date'` to it and it worked as well (although I'm working now on an error occurring in zendesk_apps_support), but it would be nice to see it solved in the official release.

nick-f:
@PolRod I've pinged the team responsible for Zendesk Apps Tools and they'll take a look at it when they can. Alternatively you can submit a PR if you have already made a fix and I'm sure that they'd appreciate that too :)

svizzari:
@PolRod Your PR has now been merged and released in v2.8.4 of Zendesk Apps Tools.

PR #227, opened 2018-02-06
Require date on check_for_update

PolRod:
Related to https://github.com/zendesk/zendesk_apps_tools/issues/223


PolRod:
Can't seem to cc @zendesk/vegemite as instructed so hope you don't mind I cc you instead, @fvzzy!

PR #233, opened 2018-03-26
sleep

ocke:
/cc @zendesk/vegemite


lib/zendesk_apps_tools/deploy.rb
```diff
@@ -2,6 +2,7 @@ module ZendeskAppsTools
   module Deploy
     def deploy_app(connection_method, url, body)
       body[:upload_id] = upload(options[:path]).to_s
+      sleep 2 # Because Aurora DB needs time to replicate
```
danielbreves: Remove the thing that should not be mentioned.

PR #234, opened 2018-03-26
skip message when flag set

ocke:
/cc @zendesk/vegemite 

We can't update CFA currently: https://samson.zende.sk/projects/conditional_fields_app/deploys/454084

skip message when flag set

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -34,7 +34,7 @@ module ZendeskAppsTools
                        hide: true,
                        desc: 'Create a version 1 app template (Deprecated)'
     def new
-      deprecated_message('error', '1.0') if options[:v1]
+      run_deprecation_checks('error', '1.0') if options[:v1]
 
       enter = ->(variable) { "Enter this app author's #{variable}:\n" }
       invalid = ->(variable) { "Invalid #{variable}, try again:" }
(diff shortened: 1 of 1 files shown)


lib/zendesk_apps_tools/command.rb
```diff
@@ -66,6 +66,10 @@ def new
 
     desc 'validate', 'Validate your app'
     shared_options(except: [:unattended])
+    method_option :'no-deprecate', type: :boolean,
```
svizzari: `--override-deprecate` or `--skip-deprecate`?

Could also have special-cased `--unattended` for this, since it was added to support deployment

lib/zendesk_apps_tools/command.rb
```diff
@@ -85,7 +89,7 @@ def validate
 
       if valid
         app_package.warnings.each { |w| say w.to_s, :yellow }
-        deprecated_message
+        deprecated_message unless options[:'no-deprecate']
```
danielbreves: Can you rename this method, it seems like it would only log a message. Also, how about `bypass-deprecation`?

danielbreves: Also, maybe add a comment and create a JIRA to remove this once all official apps are migrated to v2?

lib/zendesk_apps_tools/command.rb
ocke: good idea.

lib/zendesk_apps_tools/command.rb
ocke: > Can you rename this method, it seems like it would only log a message

Actually i was wrong when you asked, it does either log or log and exit depending on the message.

danielbreves: Still don't like that it's named `deprecated_message`, perhaps `deprecation_check`?

PR #241, opened 2018-05-18
Add command to merge marketplace content into en.json

seangoedecke:
Will merge `.md` files in translations into your `en.json`, letting you write markdown-aware fields in a markdown editor without having to cram it all into one line (and escape newlines) yourself.

Also fixes a bug where if we get a `nil` when trying to get the response version, we show a warning (we used to just silently ignore it unless the request itself failed).

FYI: drops support for Ruby 2.0, since it's incompatible with the nokogiri version we need to use for security reasons. Do not merge until we figure out if we need to communicate this breaking change.

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -316,14 +316,15 @@ module ZendeskAppsTools
         say_status 'info', 'Checking for new version of zendesk_apps_tools'
         response = Net::HTTP.get_response(URI('https://rubygems.org/api/v1/gems/zendesk_apps_tools.json'))
 
-        latest_version = Gem::Version.new(JSON.parse(response.body)["version"])
+        latest_version = Gem::Version.new(JSON.parse(response.body).fetch('version'))
+
         current_version = Gem::Version.new(ZendeskAppsTools::VERSION)
 
         cache.save 'zat_latest' => latest_version
         cache.save 'zat_update_check' => Date.today
 
         say_status 'warning', 'Your version of Zendesk Apps Tools is outdated. Update by running: gem update zendesk_apps_tools', :yellow if current_version < latest_version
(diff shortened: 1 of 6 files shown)


lib/zendesk_apps_tools/command.rb
```diff
@@ -118,6 +118,36 @@ def package
...
+          else
+            say_status 'merge_markdown', "You already have a value for #{translation_key} in en.json. Please remove it or use the -o flag", :red
+            return false
+          end
+        end
+        en_json['app'][translation_key] = File.read(file_name)
+      end
+
+      say_status 'merge_markdown', "Writing to en.json"
+      File.write("#{app_dir}/translations/en.json", en_json.to_json)
```
ocke: So `en.json` is *not* the source of truth for zendesk apps, `en.yml` is. If anybody @ zendesk would use this, it would cause problems.

However for external devs `en.json` is the source of truth, so that will work fine.

Can you make it so that it checks that if a `en.yml` is present, it writes it to `en.yml`?

seangoedecke: I don't want to make it too magical. Would you be OK with adding a flag instead to write to a yml file?

seangoedecke: @ocke I don't want to do this. The formatting differences between en.json and the yml files make the code really fiddly really quickly. If you're using a yml file you don't need this utility at all, since you can just type markdown with real newlines into a yml file with no problems (see the Text app).

seangoedecke: Discussed in 🍖. I'm gonna leave it out.

seangoedecke:
Closing this very old PR. We can reopen it if we ever figure out what's the deal with ZAT supporting old Ruby versions.

PR #243, opened 2018-06-06
Add assets_dir option for the server

rmarscher:
This adds a `--assets_dir http://proxy-server/` option for the server which enables fronting zat server with a different proxy server.

I was curious if I could use webpack and webpack dev server to develop Zendesk apps. I setup webpack dev server to listen at http://localhost:4567 and then ran `zat server --port 4568` and added a proxy config to webpack to proxy requests to http://localhost:4568.

However, the proxied app.js file was returning http://localhost:4568/ as the assetUrlPrefix which then caused the loaded app to not be the one from webpack. Opening up the assets_dir to be configured solves this problem.

See https://github.com/rmarscher/zendesk-app-webpack-example for a complete example. Thank you.


ocke:
Why have zat listen on a different port? Instead of a webpack devserver, you run `webpack watch`, which builds the files into a directory (let's say `dist/`), and then read those with zat using `zat server -p dist`.

rmarscher:
@ocke I agree that `webpack watch` is a pretty good workflow. Using webpack dev server this way enables hot reloading. I think that's the main reason for it. I actually created my example before I knew the zendesk/app_scaffold project existed. Feel free to close if you would prefer to not support it. Thanks.

hoangtrvu:
I'm closing this PR due to low activity and other workflow has been identified, thanks.

PR #244, opened 2018-06-07
Add possibility to lookup in subfolders

augustocravosilva:
Small change to lookup Theming templates in subfolders too (only 1 level deep).

@zendesk/delta @zendesk/vegemite

### DIFF
--- a/lib/zendesk_apps_tools/theme.rb
+++ b/lib/zendesk_apps_tools/theme.rb
@@ -88,10 +88,11 @@ module ZendeskAppsTools
 
       def generate_payload
         payload = {}
-        templates = Dir.glob(theme_package_path('templates', '*.hbs'))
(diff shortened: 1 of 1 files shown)


ilkkao:
looks good to me from theming point of view :+1: when green

lib/zendesk_apps_tools/theme.rb
```diff
@@ -88,10 +88,11 @@ def start_listener(callbacks_after_upload)
 
       def generate_payload
         payload = {}
-        templates = Dir.glob(theme_package_path('templates', '*.hbs'))
+        templates = Dir.glob(theme_package_path('templates', '*.hbs')) + Dir.glob(theme_package_path('templates', '*', '*.hbs'))
```
bquorning: I was just wondering, why not use an explicit `/` instead of passing multiple arguments to `theme_package_path`? In line 94 you are matching against `templates/` which includes the forward slash…

augustocravosilva: true, changed

PR #246, opened 2018-06-14
ZD3725488 Exit early when app ID is invalid

gfontanilla:
When the app ID provided in `.zat` is invalid and `zat update` is performed, the user gets a `RecordNotFound` (or `Internal Server Error` as reported in the Z1 ticket) error, which is unhelpful.

This PR adds a check to see if the app is installed, and bails early if otherwise.

@zendesk/sustaining @zendesk/vegemite 

### References

* https://support.zendesk.com/agent/tickets/3725488

### Risks

* Low: Might break private app updates via `zat update`

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -191,10 +191,12 @@ module ZendeskAppsTools
       @command = 'Update'
 
       app_id = cache.fetch('app_id') || find_app_id
-      unless /\d+/ =~ app_id.to_s
+      app_url = "/api/v2/apps/#{app_id}.json"
+      unless /\d+/ =~ app_id.to_s && check_app_installed(app_url)
         say_error_and_exit "App id not found\nPlease try running command with --clean or check your internet connection"
       end
-      deploy_app(:put, "/api/v2/apps/#{app_id}.json", {})
+
+      deploy_app(:put, app_url, {})
     end
 
     desc 'migrate', 'Helps with the migration of a v1 app to v2'
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -192,7 +192,7 @@ module ZendeskAppsTools
 
       app_id = cache.fetch('app_id') || find_app_id
       app_url = "/api/v2/apps/#{app_id}.json"
-      unless /\d+/ =~ app_id.to_s && check_app_installed(app_url)
+      unless /\d+/ =~ app_id.to_s && app_exists?(app_id)
         say_error_and_exit "App id not found\nPlease try running command with --clean or check your internet connection"
       end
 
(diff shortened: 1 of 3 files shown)


lib/zendesk_apps_tools/deploy.rb
```diff
@@ -17,6 +17,15 @@ def deploy_app(connection_method, url, body)
       say_error_and_exit e.message
     end
 
+    def check_app_installed(url)
```
danielbreves: Considering this method is simply making a GET request with the given url, it doesn't seem to be named appropriately. How about having the API path defined here and only passing the `app_id` as an argument?

lib/zendesk_apps_tools/command.rb
```diff
@@ -191,10 +191,12 @@ def update
       @command = 'Update'
 
       app_id = cache.fetch('app_id') || find_app_id
-      unless /\d+/ =~ app_id.to_s
+      app_url = "/api/v2/apps/#{app_id}.json"
```
danielbreves: This API would indicate whether the app is [available for the account](https://developer.zendesk.com/rest_api/docs/core/apps#get-information-about-app), no? To determine if there's an installation we'd need to GET [`/api/v2/apps/installations/{id}.json`](https://developer.zendesk.com/rest_api/docs/core/apps#show-app-installation).

gfontanilla: @danielbreves 

> This API would indicate whether the app is available for the account, no?

Yes :)

A request to `/api/v2/apps/installations/` and searching for the app ID in the result would work as well. 

If the app is installed, the endpoint `/api/v2/apps/{app_id}.json` should exist/return a 200 (404 if not). The method employed in this PR makes us of this assumption, which I think is quicker (no need to search through a list of IDs).

fbvilela: yes, Daniel is correct here.

lib/zendesk_apps_tools/deploy.rb
fbvilela: agree, this method will make more sense once it loops through the installations trying to find the `app_id` which should be passed in.

lib/zendesk_apps_tools/command.rb
fbvilela: fair enough, if you no longer care if it's installed or not, that endpoint will work 👍

fbvilela: ...as long as you make the request using the account's subdomain and being authenticated, off course

fbvilela: not a caching issue... this endpoint will return the app IF the app is available to you. that's what you want, right? to know if the id exists? What if you are trying to `zat update` an app that hasn't been installed yet? I think you can trust that endpoint. The only problem I see is if you have a wrong id in your .zat file which matches another existing app_id from an app that belongs to another account. In this case, the endpoint will NOT return 200 although the app would exist.

PR #247, opened 2018-06-15
Fix build

libo:
`https://rubygems.org/api/v1/gems/zendesk_apps_tools.json` stub was missing `version` and this error was generated:

```
../home/travis/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/site_ruby/2.0.0/rubygems/version.rb:209:in `initialize': Malformed version number string  (ArgumentError)
```

Probably has been working in the past based due to the random order of the specs and their stubs

RISK: low, fix a spec


libo:
cc: @augustocravosilva @svizzari @ocke

PR #196, opened 2017-07-13
[APPS-3473] Multi Product - Use v3 app market endpoints and install in multiple products

fbvilela:
#:bear:

/cc @zendesk/wombat
@zendesk/vegemite 

### Description
This is part of the multi product effort
`zat create` to install the app in all product locations found in manifest.
Depends on the ZAM PR listed below

### References
* JIRA: https://zendesk.atlassian.net/browse/APPS-3473
* ZAM PR: https://github.com/zendesk/zendesk_app_market/pull/2599

### Risks
* low - breaks app creation/installation

PR #250, opened 2018-06-28
[AF-1067] Only return uniq product_codes

Mehul-Porwal:
When we execute `zat create` the app is installed multiple times(number of location times) per product.

This is caused due to a minor error in logic here: 
```
manifest.location_options.collect{ |option| option.location.product_code }
```

Consider an example where the manifest looks something like this:
```
"location": {
    "support": {
      "ticket_sidebar":"http://localhost:44000/parent",
      "modal": "http://localhost:44000/modal"
    }
  },
```

Now based on the above logic, we are transforming an array of 2 element like this:
`
[
{'location':'ticket_sidebar', 'product_code': 1, .... },
{'location':'modal', 'product_code': 1, .... }
]
` ==> `[1,1]`

The fix is just to return uniq values.

@zendesk/vegemite 

## Risk
[LOW] There is only one consumer for the `product_codes` function that is modified in this PR, that is `zat create` command. This command is already broken, which is being fixed in this PR.

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -248,7 +248,7 @@ module ZendeskAppsTools
     end
 
     def product_codes(manifest)
-      manifest.location_options.collect{ |option| option.location.product_code }
+      manifest.location_options.collect{ |option| option.location.product_code }.uniq
     end
 
     def settings


danielbreves:
Looks like this was introduced in https://github.com/zendesk/zendesk_apps_tools/pull/196 to support installing the app in multiple products.

Mehul-Porwal:
@hoangtrvu No tests are written for this ATM, but since we are under production freeze, I can take this time to add some tests for this function.

token-cjg:
A test in this gem to cover this case would be useful.

PR #257, opened 2018-09-27
Replace 2.4 method match? with 2.0 compatible method match

yu-tian113:
:v:

/cc @zendesk/vegemite
@zendesk/apps-migration 

### Tasks
- [x] Replace Ruby 2.4 method `match?` with `match` to support older Ruby versions (v2.0.0+)

### Risks
* [low] `zat new --scaffold` should not break

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    zendesk_apps_tools (2.10.0)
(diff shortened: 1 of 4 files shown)


token-cjg:
Could you add a test or two to `spec/lib/zendesk_apps_tools/command_spec.rb` for the case when passing an optional `--scaffold` parameter to the `zat new` command?

PR #259, opened 2018-10-24
Remove `zat migrate` and all associated options and tests from ZAT

yu-tian113:
:v:

/cc @zendesk/vegemite
@zendesk/apps-migration 

### Tasks
- [x] Remove `zat migrate` command and associated methods
- [x] Remove tests for `zat migrate`

### References
* JIRA: https://zendesk.atlassian.net/browse/OFFAPPS-1066

### Risks
* [low] Complete remove `zat migrate` command and tests, without touching any of other functions.


vaevictis:
I was just thinking, we don't have other tools assuming this command exists? I don't believe much would be relying on it, but if it's the case shouldn't we do some kind of deprecation warning?

yu-tian113:
good point, not that I know of, to be confirmed with vegemite.

PR #263, opened 2018-11-28
Delete Cache if ZAT commands or requests are unsuccessful

keitheous:
:v:

/cc @zendesk/vegemite

### Description
When running zat commands, the system prompts for developers to key in the subdomain and email address. If these credentials are invalid, the system will cache these credentials and reuse them in future attempts to reconnect.

As a solution, we delete cache whenever there's a connection error. We remove subdomain, username, and app_id from .zat file in the local directory. 

### Tasks
- [x] create a method `Cache#delete` to delete specific cache (i.e. username, subdomain, app_id)
- [x] write specs for `Cache#delete`
- [x] delete cache for username, subdomain, and app_id when necessary
- [x] refactor module `ZendeskAppsTools::Deploy`
- [x] write test for `ZendeskAppsTools::Deploy.find_app_id` for refactoring
- [x] Refactor `ZendeskAppsTools::Deploy` further

### Implementation Notes
<details>
<summary>Better way to write `.empty?` condition</summary>
better way to write empty condition
</details>

```
response_json = ""

app_json =  response_json.empty? ? nil : 'method_name' <--(!!!)

app_json = 'method_name' unless response_json.empty? <---(Better way)

(because app_json will be nil otherwise anyway)
```

<details>
<summary>Extracting get_connection into a private method</summary>
making connection a private function
</details>
<br>
<details>
<summary>My discussion with Dan Breves: Clear Cache occurs whenever the code goes into a rescue block.</summary>
This rescue block is designed to not crash from connection/authentication errors. Which means that when there's a problem with the internet and the system fails to connect, this cache will clear.
 
The next time a developer tries to reconnect, the developer is required to key in their username, subdomain and password again.
Typically when everything works fine ~(with awesome internet)~, they only have to key in their passwords
</details>

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1433

### Risks
* [low?] Refactored Module ZendeskAppsTools::Deploy - developers might not be able to deploy effectively


lib/zendesk_apps_tools/deploy.rb
```diff
@@ -112,7 +118,18 @@ def check_job(job_id)
...
+
+    private
+
+    def delete_cache_exit_and_error(error_message)
+      cache.delete(['app_id', 'subdomain', 'username'])
+      say_error_and_exit error_message
+    end
+
+    def connection(multipart = nil)
+      get_connection multipart
```
danielbreves: Why not use `get_connection` directly everywhere?

keitheous: As discussed in real life. get_connection called with .post .get doesn't promote readability. 
(i.e. `get_connection.get('url')`)

The other alternative that we could use is ruby alias method name. We could call 
```
  alias_method 'name', 'full_name'
```
which is essentially doing what the method is doing.

lib/zendesk_apps_tools/api_connection.rb
```diff
@@ -6,7 +6,7 @@ module APIConnection
     def prepare_api_auth
       @subdomain ||= cache.fetch('subdomain') || get_value_from_stdin('Enter your Zendesk subdomain or full Zendesk URL:')
       @username  ||= cache.fetch('username', @subdomain) || get_value_from_stdin('Enter your username:')
-      @password  ||= cache.fetch('password', @subdomain) || get_password_from_stdin('Enter your password:')
+      @password  ||= get_password_from_stdin('Enter your password:')
```
keitheous: password is not stored in `.zat`

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -112,7 +118,18 @@ def check_job(job_id)
...
       end
     rescue Faraday::Error::ClientError => e
-      say_error_and_exit e.message
+      delete_cache_exit_and_error(e.message)
+    end
+
+    private
+
+    def delete_cache_exit_and_error(error_message)
+      cache.delete(['app_id', 'subdomain', 'username'])
```
danielbreves: Should we output a message saying we're deleting the user cache because of the failure?

spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      def say_status(command, text)
+      end
+
+      def get_value_from_stdin(text)
+        'app_2'
+      end
+
+      def say(message, color = nil)
+      end
```
danielbreves: Could we replace most of these with ` allow(subject).to receive_message`?

lib/zendesk_apps_tools/deploy.rb
fvzzy: I like the `alias_method` suggestion

lib/zendesk_apps_tools/api_connection.rb
ocke: it can be.

lib/zendesk_apps_tools/deploy.rb
fvzzy: @danielbreves seems a bit overkill. Why would you want it to store potentially incorrect user inputs? As a developer/user of this tool I'm not sure I'd care the cache is being deleted but rather than they're just cached after a successful connection has been established once

lib/zendesk_apps_tools/api_connection.rb
keitheous: Okay Olaf

.github/pull_request_template.md
```diff
@@ -13,7 +13,7 @@ Describe the original problem and the changes made on this PR.
 * JIRA: https://zendesk.atlassian.net/browse/AF-XXX
 
 ### Risks
-* [HIGH | medium | low] Does it work across browsers (including IE!)?
```
fvzzy: As these changes are unrelated, it's preferable to split them out into their own PR (even if that PR is really small).

As an aside, "Does it work in the different products (Support, Chat, Connect)?" is probably also not relevant to ZAT

keitheous: Sounds good

spec/lib/zendesk_apps_tools/deploy_spec.rb
keitheous: Yes we can

ocke:
I'm not sure if we should delete these keys in the `.zat` file when it fails. What if I'm doing a `zat update` and my internet is disconnected? Boom, now my `.zat` file empty :|

I'm a bit confused on the functionality of this and why we need this. Can't people just simply edit their zat file when they need to? We might communicate a bit better about .zat, but not sure if I like clearing it.

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -87,23 +94,22 @@ def check_status(response, poll_job = true)
...
+        response = check_job_connection.get("/api/v2/apps/job_statuses/#{job_id}")
         info     = json_or_die(response.body)
         status   = info['status']
-        message  = info['message']
-        app_id   = info['app_id']
 
         if %w(completed failed).include? status
           case status
           when 'completed'
+            app_id   = info['app_id']
```
keitheous: whitespace

spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      subject = subject_class.new(mock_response.new(''))
+
+      expect(subject).to receive(:say_error).with(
+        "App not found. " \
+        "Please verify that your credentials, subdomain, and app name are correct."
+      )
+      expect { subject.find_app_id }.to raise_error(SystemExit)
+    end
+
+    it 'returns app id if is in response_with_apps' do
```
keitheous: returns app id if app exist in response_with_apps

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -1,10 +1,13 @@
+require 'faraday'
+require 'zendesk_apps_tools/common'
+
```
keitheous: explicitly require external dependency zendeskappstools/api_connection

include Z::A

keitheous:
After a discussion with Farhan, Mehul and Olaf IRL, we have decided to try and explore a different approach.

lib/zendesk_apps_tools/deploy.rb
keitheous: I'll use alias_method :)

keitheous:
Closed and split into 2 separate PRs

1) https://github.com/zendesk/zendesk_apps_tools/pull/265
2) TBC - improved approach of handle cached ZAT credentials

PR #265, opened 2018-12-02
Refactor find_app_id in ZendeskAppsTools::Deploy

keitheous:
:v:

/cc @zendesk/vegemite

### Description
While working on a sprint task, I came across some ruby code that doesn't look great. Thought I'd clean it up abit before working on that sprint task. This change follows the [Ruby Style Guide](https://github.com/rubocop-hq/ruby-style-guide).

Prior to the refactoring, I created a test for the code that is to be refactored. Then made sure that the test passes after refactoring

### Tasks
1. Write Test for method
2. Refactor method
3. Run test for new method

4. Explicitly declare external dependencies (modules)

### Implementation Notes
1. Better way to write `.empty?` condition
  ```
    response_json = ""

    app_json =  response_json.empty? ? nil : 'method_name' <--(!!!)

    app_json = 'method_name' unless response_json.empty? <---(Better way)

    (because app_json will be nil otherwise anyway)
  ```

2. get_connection was extracted into a private method, and then converted into alias_method

### References
* [StackExchange: Why write tests for code that I will refactor?](https://softwareengineering.stackexchange.com/questions/233222/why-write-tests-for-code-that-i-will-refactor)
* [Alias Method](https://blog.bigbinary.com/2012/01/08/alias-vs-alias-method.html) 
* [Original PR](https://github.com/zendesk/zendesk_apps_tools/pull/263)
* [ Ruby Style Guide](https://github.com/rubocop-hq/ruby-style-guide)

### Risks
* [low] Refactored method by first creating the test, then refactor code to pass test.

### DIFF
--- a/lib/zendesk_apps_tools/deploy.rb
+++ b/lib/zendesk_apps_tools/deploy.rb
@@ -100,10 +100,10 @@ module ZendeskAppsTools
     end
 
     def check_job(job_id)
-      check_job_connection = connection
+      http_client = connection
 
       loop do
-        response = check_job_connection.get("/api/v2/apps/job_statuses/#{job_id}")
+        response = http_client.get("/api/v2/apps/job_statuses/#{job_id}")
         info     = json_or_die(response.body)
         status   = info['status']
 
--- a/lib/zendesk_apps_tools/deploy.rb
+++ b/lib/zendesk_apps_tools/deploy.rb
@@ -1,9 +1,16 @@
+require 'faraday'
+require 'zendesk_apps_tools/common'
+require 'zendesk_apps_tools/api_connection'
+
 module ZendeskAppsTools
   module Deploy
+    include ZendeskAppsTools::Common
+    include ZendeskAppsTools::APIConnection
+    alias_method :connection, :get_connection
+
     def deploy_app(connection_method, url, body)
       body[:upload_id] = upload(options[:path]).to_s
       sleep 2 # Because the DB needs time to replicate
-      connection = get_connection
 
       response = connection.send(connection_method) do |req|
         req.url url
@@ -19,7 +26,6 @@ module ZendeskAppsTools
 
     def app_exists?(app_id)
       url = "/api/v2/apps/#{app_id}.json"
-      connection = get_connection
       response = connection.send(:get) do |req|
         req.url url
       end
@@ -28,7 +34,6 @@ module ZendeskAppsTools
     end
 
     def install_app(poll_job, product_name, installation)
-      connection = get_connection
       response = connection.post do |req|
         req.url "api/#{product_name}/apps/installations.json"
         req.headers[:content_type] = 'application/json'
@@ -38,7 +43,6 @@ module ZendeskAppsTools
     end
 
     def upload(path)
-      connection = get_connection :multipart
       zipfile_path = options[:zipfile]
 
       if zipfile_path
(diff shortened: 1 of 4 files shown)


spec/lib/zendesk_apps_tools/command_spec.rb
```diff
@@ -144,7 +144,10 @@
         stub_request(:get, PREFIX + '/api/apps.json')
           .with(headers: { 'Authorization' => 'Basic dXNlcm5hbWU6cGFzc3dvcmQ=' })
           .to_return(body: '')
-        expect(@command).to receive(:say_error).with(/^The app was not found./)
+        expect(@command).to receive(:say_error).with(
```
fvzzy: Food for thought for test cases like this: do you think it might be worth relaxing the check here, so that if the error message changes the test continues to pass? i.e. How strict should we be in what you would you consider an actual test "failure" here?

(Not at all a blocker btw)

spec/lib/zendesk_apps_tools/cache_spec.rb
```diff
@@ -6,12 +6,13 @@
...
-    let(:cache) { ZendeskAppsTools::Cache.new(path: tmpdir) }
-    let(:tmpdir) { Dir.mktmpdir }
+    let(:tmpdir)   { Dir.mktmpdir }
+    let(:cache)    { ZendeskAppsTools::Cache.new(path: tmpdir) }
+    let(:zat_file) { File.join(tmpdir, '.zat') }
 
     before do
       content = JSON.dump(subdomain: 'under-the-domain', username: 'Roger', app_id: 12)
-      File.write(File.join(tmpdir, '.zat'), content, mode: 'w')
+      File.write(zat_file, content, mode: 'w')
```
fvzzy: Rather than actually writing to the file system for this test, would it be sufficient (and beneficial?) to instead write to an object like a [StringIO](https://ruby-doc.org/stdlib-2.5.1/libdoc/stringio/rdoc/StringIO.html) here?

fvzzy:
See minor comments, otherwise 👍

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -87,22 +100,21 @@ def check_status(response, poll_job = true)
     end
 
     def check_job(job_id)
-      connection = get_connection
+      check_job_connection = connection
```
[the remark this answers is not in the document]
keitheous: Hmm.. probably not necessary. The reason why `connection` is assigned to `check_job_connection` here, is because of the loop that follows (line 105). 

`connection` which is essentially `get_connection` (from APIConnection Module) goes through the process of preparing network client and fetching credentials / prompting for user inputs. I didn't to want to spam the process in the loop.

as for using `alias_method :connection, :get_connection`,
it provides a DRYER option where we don't have to repeat `connection = get_connection` in all the methods of this Module.

spec/lib/zendesk_apps_tools/command_spec.rb
keitheous: I love food (including for thoughts). I now understand how they were using /Regex/ to check for this failure. Thanks for pointing this out.

Before the change, the Error Message was 
`'The app was not found. Please verify your credentials, subdomain, and app name are correct.' ` and they were checking for `/^The app was not found./`. If I wanted to relax the check here it would have been

`"/^App not found./` for the current
`"App not found. Please verify that your credentials, subdomain, and app name are correct."`

I usually prefer to be explicitly declare things for less brain processing. But I am also interested to know what you think? When should I consider relaxing the checks here.

lib/zendesk_apps_tools/deploy.rb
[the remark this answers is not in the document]
keitheous: When discuss in real life, we came up with a more appropriate name for variable `check_job_connection`. We should name it what connection represents before the loop block. I will rename this as `network_client`

PR #267, opened 2018-12-12
Optimize network client creation

keitheous:
/cc @zendesk/vegemite :v:

### Description
Found something to optimize in ZAT. We use Faraday Gem for our HTTP network client.

This is set up by the `APIConnection` module, specifically in the `get_connection` method. `get_connection` relies on another method called `prepare_api_auth` within the same module. 

`prepare_api_auth` prompts user for subdomain, username and password if none of these information are available in zat cache. `get_connection`, on the other hand, uses these credentials to build the network client.

## Implementation Notes
`get_connection` is called many times in `deploy.rb`. Every single `get_connection` **in a single run time** creates a new network client (faraday object) and reiterates through `prepare_api_auth` to reassign subdomain, username, password to themselves. Let's dive deeper into code!
<details>
<summary>The Problem?</summary>
<br/>
<em>we puts as indicators to understand the flow of execution</em>
screen shot 2018-12-12 at 9 44 14 am

<br/><em>upon running zat create with input credentials, we see 4 get_connections and 4 prepare auths.</em>
screen shot 2018-12-12 at 9 42 10 am

<br/><em>every time we go through the Faraday.new block, I had a sneaky suspicion that we were creating new Faraday Objects for deploy.md to make http request. Hence more indicators</em>
screen shot 2018-12-12 at 9 56 40 am

<br/><em>Network Clients Clones</em>
screen shot 2018-12-12 at 9 55 53 am

<em>Out of the four objects that were created, I have identifed that 2 were necessary. The first is for network client with url encoding, the second - for network client with multipart encoding. Three and four are just clones of the first. </em>
</details>

#### The solution

1) Inside `get_connection`, prevent `prepare_api_auth` if you already have the subdomain, username and password. In other words, execute `prepare_api_auth` unless we have @subdomain or @username or @password

    ```
    (api_connection.rb)

    def get_connection
      ...
      prepare_api_auth unless @subdomain && @username && @password
      ...(go faraday)..
    end
    ```

2) In `deploy.rb`, whenever we call `connection.send(url)`, `connection.get(url)`, or `connection.post(url)`, use the same connection. The method below caches connection in a local variable, after calling it once, which is accessible throughout `deploy.rb` in a single run time.

    ```
    (deploy.rb)

    def connection
      @connection ||= get_connection
    end
    ```
`get_connection(:multipart)` is only called once, so let's call get_connection directly on that one.

<details>
<br/>
<summary>The Result?</summary>
<em>running zat create once again, we can see that there's much of an improvement.</em><br/>
screen shot 2018-12-12 at 10 52 39 am<br/><em>
Our first `get_connection` produced a faraday object with multipart url encoding. The second would be our regular one that is being recycled throughout `deploy.rb`. `prepare_api_auth` occurred just once.</em>
</details> 

### Tasks
- [x] prevent iterating through `prepare_api_auth ` if we already have subdomain, username, and password.
- [x] write test for `get_connection` to identify how many times `prepare_api_auth` is called.
- [x] prevent creating multiple/new network client in every `deploy.md` methods, unless necessary (multipart url encoding)

### References
* no references

### Risks
* [low] - Optimize network client by reusing credentials and client that has already been defined.


lib/zendesk_apps_tools/deploy.rb
```diff
@@ -126,5 +123,11 @@ def check_job(job_id)
     rescue Faraday::Error::ClientError => e
       say_error_and_exit e.message
     end
+
+    private
+
+    def connection
```
fvzzy: Might there be a less ambiguous name for this method, e.g. `cached_connection`?

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -56,7 +55,7 @@ def upload(path)
         uploaded_data: Faraday::UploadIO.new(package_path, 'application/zip')
       }
 
-      response = connection(:multipart).post('/api/v2/apps/uploads.json', payload)
+      response = get_connection(:multipart).post('/api/v2/apps/uploads.json', payload)
```
danielbreves: I think it may be unclear when reading this why we sometimes use `connection` and other times `get_connection`. What if we kept a cache keyed by the encoding?
e.g.
```
def cached_connection(encoding = :url_encoded)
  @connection ||= {}
  @connection[encoding] ||= get_connection(encoding)
end
```

keitheous: wow this is pretty cool. 

1) Say we get a `cached_connection(:url_encoded)` the first time,
* the first line becomes `@connection = {}`
* the second line becomes `@connection[:url_encoded] = get_connection(:url_encoding)`

2) The next time someone calls cache_connection(:url_encoded),
* the first line evaluates to just `@connection` because it is already defined
* the second line returns `@connection[:url_encoded]` which was assigned previously

3) The next next time someone calls cache_connection(:multipart),
* the first line evaluates to just`@connection` again because it is already defined
* however the second line becomes `@connection[:multipart] = get_connection(:multipart)`

Now it has  both values cached - Brilliant! Thanks, @danielbreves

lib/zendesk_apps_tools/deploy.rb
keitheous: Yeah let's do that 👍

lib/zendesk_apps_tools/deploy.rb
keitheous: screen shot 2018-12-13 at 10 25 43 pm

spec/lib/zendesk_apps_tools/api_connection_spec.rb
```diff
@@ -134,4 +136,33 @@ def initialize(subdomain = nil, username = nil)
       end
     end
   end
+
+  describe '#get_connection' do
+    context 'with missing credentials' do
+      it 'calls prepare_api_auth for password when password is missing' do
+        subject = subject_class.new('subdomain', 'username')
+        expect(subject).to receive(:prepare_api_auth).exactly(1).times
```
fvzzy: Is it possible to narrow the scope of this test (and the one below) so that we're testing it is `received` with certain parameters?

spec/lib/zendesk_apps_tools/api_connection_spec.rb
```diff
@@ -134,4 +136,33 @@ def initialize(subdomain = nil, username = nil)
...
+        subject.get_connection
+      end
+    end
+
+    context 'with all credentials' do
+      let(:subject) { subject_class.new('subdomain', 'username', 'password') }
+
+      it 'does not call prepare_api_auth for credentials' do
+        expect(subject).to receive(:prepare_api_auth).exactly(0).times
```
fvzzy: `.to_not receive(prepare_api_auth)` if it works?

spec/lib/zendesk_apps_tools/api_connection_spec.rb
```diff
@@ -134,4 +136,33 @@ def initialize(subdomain = nil, username = nil)
...
+        subject.get_connection
+      end
+
+      it 'creates a faraday network client for requests' do
+        expect(subject.get_connection.class).to eq(Faraday::Connection)
```
fvzzy: Untested, but maybe `.to be_instance_of(Faraday::Connection)` if it works here?

fvzzy:
Minor comments, otherwise 👍

spec/lib/zendesk_apps_tools/api_connection_spec.rb
keitheous: Hi @fvzzy , If you have sometime today, I might grab you in person to elaborate more on this?

spec/lib/zendesk_apps_tools/api_connection_spec.rb
keitheous: Of course. I forgot about `be_instance_of`

spec/lib/zendesk_apps_tools/api_connection_spec.rb
keitheous: Yes it does! :)

spec/lib/zendesk_apps_tools/api_connection_spec.rb
keitheous: As discussed in real life, I will rename the two tests so that it reads better. 👍

keitheous:
@Mehul-Porwal 

It is not necessary. It went up due to an unrelated issue. I am working on a [fix](https://github.com/zendesk/zendesk_apps_tools/pull/270) for that. 

In short, this is due to a missing patch version in one of the dependencies in gemspec (gem listen) - whenever we bump, the other dependencies of gem gets updated as well. In long, we will discuss this in person

PR #268, opened 2018-12-13
Fix email validation

keitheous:
/cc @zendesk/vegemite :v:

### Description
After adding validation for username which is end user's email. I have not considered for users who uses API token. 

### Implementation Notes
The previous regular expression for email was
```
EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})$/i
```

This is now changed to 
```
EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})(\/token:\S+)?$/i
```

screen shot 2018-12-13 at 3 01 26 pm

### Tasks
- [x] Test email with `email@email.email/token:{api_token}`

### References
* https://zendesk.slack.com/archives/C1230V1CG/p1544631643154400

### Risks
* [low] Attempt to fix ZAT login with email that requires api_token

### DIFF
--- a/lib/zendesk_apps_tools/api_connection.rb
+++ b/lib/zendesk_apps_tools/api_connection.rb
@@ -4,7 +4,7 @@ module ZendeskAppsTools
     # taken from zendesk/lib/vars.rb
     SUBDOMAIN_VALIDATION_PATTERN = /^[a-z0-9][a-z0-9\-]{1,}[a-z0-9]$/i
     ZENDESK_URL_VALIDATION_PATTERN = /^(https?):\/\/[a-z0-9]+(([\.]|[\-]{1,2})[a-z0-9]+)*\.([a-z]{2,16}|[0-9]{1,3})((:[0-9]{1,5})?(\/?|\/.*))?$/ix
-    EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})(\/token:\S+)?$/i
+    EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})(\/token)?$/i
 
     EMAIL_ERROR_MSG = 'Please enter a valid email address.'
     PROMPT_FOR_URL  = 'Enter your Zendesk subdomain or full URL (including protocol):'
--- a/lib/zendesk_apps_tools/api_connection.rb
+++ b/lib/zendesk_apps_tools/api_connection.rb
@@ -4,7 +4,7 @@ module ZendeskAppsTools
     # taken from zendesk/lib/vars.rb
     SUBDOMAIN_VALIDATION_PATTERN = /^[a-z0-9][a-z0-9\-]{1,}[a-z0-9]$/i
     ZENDESK_URL_VALIDATION_PATTERN = /^(https?):\/\/[a-z0-9]+(([\.]|[\-]{1,2})[a-z0-9]+)*\.([a-z]{2,16}|[0-9]{1,3})((:[0-9]{1,5})?(\/?|\/.*))?$/ix
-    EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})$/i
+    EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})(\/token:\S+)?$/i
 
     EMAIL_ERROR_MSG = 'Please enter a valid email address.'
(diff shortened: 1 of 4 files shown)


lib/zendesk_apps_tools/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module ZendeskAppsTools
-  VERSION = '2.12.0'
+  VERSION = '2.12.1'
```
keitheous: This was left out during the previous bump

spec/lib/zendesk_apps_tools/api_connection_spec.rb
```diff
@@ -2,11 +2,32 @@
...
+      expect(matches_email?('u!sername@email.com')).to eq(false)
+    end
+
+    it 'will match valid email addresses' do
+      expect(matches_email?('username@email.com')).to eq(true)
+      expect(matches_email?('username123@e-mail.com')).to eq(true)
+    end
+
+    it 'will match valid email addresses with api_token' do
+      expect(matches_email?('username@email.com/token:p0K3Mon')).to eq(true)
```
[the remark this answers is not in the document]
keitheous: Love the enthusiasm!

fvzzy:
Could be slightly more specific with the token regex, but otherwise looks fine 👍

spec/lib/zendesk_apps_tools/api_connection_spec.rb
```diff
@@ -2,11 +2,32 @@
 require 'api_connection'
 
 describe ZendeskAppsTools::APIConnection do
-  let(:subdomain_validation_pattern) { ZendeskAppsTools::APIConnection::SUBDOMAIN_VALIDATION_PATTERN }
-  let(:url_validation_pattern)       { ZendeskAppsTools::APIConnection::ZENDESK_URL_VALIDATION_PATTERN }
-  let(:default_url_template)         { ZendeskAppsTools::APIConnection::DEFAULT_URL_TEMPLATE }
+  describe 'EMAIL_REGEX' do
```
vaevictis: Nice extra set of tests! ❤️

token-cjg:
Actually, should we instead have email of the form

`username@email.com/token` and then password as `p0K3Mon`?  I'm having difficulties testing this locally.

lib/zendesk_apps_tools/api_connection.rb
```diff
@@ -4,7 +4,7 @@ module APIConnection
     # taken from zendesk/lib/vars.rb
     SUBDOMAIN_VALIDATION_PATTERN = /^[a-z0-9][a-z0-9\-]{1,}[a-z0-9]$/i
     ZENDESK_URL_VALIDATION_PATTERN = /^(https?):\/\/[a-z0-9]+(([\.]|[\-]{1,2})[a-z0-9]+)*\.([a-z]{2,16}|[0-9]{1,3})((:[0-9]{1,5})?(\/?|\/.*))?$/ix
-    EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})$/i
+    EMAIL_REGEX  = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})(\/token:\S+)?$/i
```
token-cjg: You should probably drop the requirement for a string after token here

token-cjg: You should probably drop the requirement for a string after token here

hoangtrvu: As Chris said, for token scenario, the email would be `abc@example.com/token` and the token would be provided in the password field.

keitheous: Hey buddy, you're absolutely right! I am so glad you caught this. Bravo Chris.


For context, IRL Chris has brought to my attention that I might have understood the [documentations](https://developer.zendesk.com/rest_api/docs/support/requests#api-token) wrong. 

When prompted to enter credentials when doing `zat create`, after keying in subdomain, developers should be allowed to enter `full_email/token` as username and API TOKEN as password (Api Token is generated in Admin > Channels > API ).

I tested this out and here is the result.

screen shot 2018-12-13 at 9 25 13 pm

keitheous:
Regarding the gem versioning. The current version should be VERSION = '2.12.1'. This was left out.

When the branch gets merged, we should do bundle exec rake bump (VERSION = '2.12.2')

token-cjg:
For clarification, I built the gem locally and tested

1) the happy path for my sponsored account & standard email/password
2) the happy path for my sponsored account & email with token/api_token as password (now working, this was not working before)
3) the fail validation path for url
4) the fail validation path for email

PR #269, opened 2018-12-18
Relocate cache.save zat_details

keitheous:
:v:

/cc @zendesk/vegemite

### Description
In the past, when a wrong `subdomain` or `username` is entered during a `zat_update` or `zat_create`, the connection fails and these details are stored in a cache file locally. In subsequent attempts, zat will look into the local cache file to reconnect using the wrong `subdomain` or `username`. 

This is because `subdomain` and `username` are stored directly after user prompts. Instead, they should be stored after a successful connection. This successful connection occurs and is defined when API `response.body['status']` is [**completed**](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/deploy.rb#L112).

#### 1) Decision to piggyback off [cache.save 'app_id'](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/deploy.rb#L114)
Felt that this is a good place to cache [instance variables `subdomain` and `username`](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/api_connection.rb#L26). Reasons: 
- Instance Variables meant that they are accessible across modules.
- These Variables are set - after `ZendeskAppsTools::APIConnection#get_connection` has already been called ([and potentially cached](https://github.com/zendesk/zendesk_apps_tools/pull/267)) earlier on.
- This occurs within the `case when` the `request.body['status']` is [completed](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/deploy.rb#L112) block. Meaning the connection has already been established, and `username`/`subdomain` are valid.
- ALL methods in Deploy EVENTUALLY ends up in this block anyway. INCLUDING `app_exists?`, `upload` and `find_app_id`. ~Just like the river and the ocean~

#### 2) Decision to rename [info](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/deploy.rb#L107) to zat_details

<details>
<summary>Info seems to be a reserved ruby/system keyword</summary>
<em>When hash is info</em>
hash before renamed
<em>When hash is renamed to response</em>
hash after renamed

</details>

#### 3) Decision to pass the ENTIRE RESPONSE into private method zat_details as param

Essentially, private method zat_details becomes a wrapper method that returns a hash object, as such

```
cache.save sexy_hash

private

def sexy_hash
  {
    key_1: 'value_1',
    key_2: 'value_2',
    key_3: 'value_3'
  }
end
```
which is equivalent to `cache.save 'key_1' => 'value_1', 'key_2' => 'value_2', 'key_3' => 'value_3'`. Smaller blocks and methods reads a whole lot easier. Along with this approach, I have also identified a few areas in the current script to improve in the **next PR**.

**4) Optional reading: Storing App Id or App Name to [response](https://github.com/zendesk/zendesk_apps_tools/blob/master/lib/zendesk_apps_tools/deploy.rb#L66)**
Response return a bunch of powerful information. <details> <summary>What powerful information you ask?
</summary>
reasons for passing response
</details>

With app_name, app_id and app_url, we can avoid prompting users for app_name, or even executing `find_app_id` from `/api/apps.json`  😱 ~wait till you see the response data~ We can potentially store more things in zat cache. This is implementation work for another day. 

### Screenshot
Result
screen shot 2018-12-18 at 4 09 00 pm

### Tasks
- [x] Document approach and decision.
- [x] Write Test for ~cache?~ check_job method

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1433

### Risks
* [low] Relocate cache.save `subdomain` and `username`. Cached details are only complementary.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    zendesk_apps_tools (2.12.1)
+    zendesk_apps_tools (2.12.2)
       execjs (~> 2.7.0)
       faraday (~> 0.9.2)
       faye-websocket (~> 0.10.7)
--- a/lib/zendesk_apps_tools/deploy.rb
+++ b/lib/zendesk_apps_tools/deploy.rb
@@ -129,13 +129,12 @@ module ZendeskAppsTools
     private
 
     def zat_details(response)
-      zat_details = {}
+      zat_details = {
+        'subdomain' => @subdomain,
+        'username' => @username
+      }
 
-      zat_details['app_id']    = response['app_id'] if response['app_id']
-      zat_details['subdomain'] = @subdomain
-      zat_details['username']  = @username
-
-      zat_details
+      zat_details.merge('app_id' => response['app_id']) if response['app_id']
     end
   end
 end
--- a/lib/zendesk_apps_tools/deploy.rb
+++ b/lib/zendesk_apps_tools/deploy.rb
@@ -103,18 +103,17 @@ module ZendeskAppsTools
       http_client = connection
 
       loop do
-        response = http_client.get("/api/v2/apps/job_statuses/#{job_id}")
-        info     = json_or_die(response.body)
-        status   = info['status']
+        request  = http_client.get("/api/v2/apps/job_statuses/#{job_id}")
+        response = json_or_die(request.body)
+        status   = response['status']
 
         if %w(completed failed).include? status
           case status
           when 'completed'
-            app_id = info['app_id']
-            cache.save 'app_id' => app_id if app_id
+            cache.save zat_details(response)
             say_status @command, 'OK'
           when 'failed'
-            say_status @command, info['message'], :red
+            say_status @command, response['message'], :red
             exit 1
           end
           break
@@ -125,6 +124,18 @@ module ZendeskAppsTools
       end
     rescue Faraday::Error::ClientError => e
       say_error_and_exit e.message
+    end
+
+    private
+
+    def zat_details(response)
+      zat_details = {}
+
+      zat_details['app_id']    = response['app_id'] if response['app_id']
+      zat_details['subdomain'] = @subdomain
+      zat_details['username']  = @username
+
+      zat_details
     end
   end
 end
--- a/lib/zendesk_apps_tools/deploy.rb
+++ b/lib/zendesk_apps_tools/deploy.rb
@@ -110,7 +110,7 @@ module ZendeskAppsTools
         if %w(completed failed).include? status
           case status
           when 'completed'
-            cache.save zat_details(response)
+            cache.save zat_contents(response)
             say_status @command, 'OK'
           when 'failed'
             say_status @command, response['message'], :red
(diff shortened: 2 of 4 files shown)


lib/zendesk_apps_tools/deploy.rb
```diff
@@ -126,5 +125,17 @@ def check_job(job_id)
...
+
+    def zat_details(response)
+      zat_details = {}
+
+      zat_details['app_id']    = response['app_id'] if response['app_id']
+      zat_details['subdomain'] = @subdomain
+      zat_details['username']  = @username
+
+      zat_details
+    end
```
keitheous: Might tidy this up.

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    zendesk_apps_tools (2.12.1)
+    zendesk_apps_tools (2.12.2)
```
keitheous: Waiting for [this](https://github.com/zendesk/zendesk_apps_tools/pull/267) to be approved and merged. Will remove from this commit.

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -126,5 +125,16 @@ def check_job(job_id)
...
+    private
+
+    def zat_details(response)
+      zat_details = {
+        'subdomain' => @subdomain,
+        'username' => @username
+      }
+
+      zat_details.merge('app_id' => response['app_id']) if response['app_id']
+    end
```
keitheous: Another way I can do this is.
```
def zat_details(response)
  zat_details = {}
  zat_details['subdomain'] = @subdomain
  zat_details['username'] = @username
  zat_details['app_id'] = response['app_id']) if response['app_id']

  zat_details
```

This one ☝️ might be better

fvzzy: Yeah the version in the comment seems more readable to me

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -126,5 +125,16 @@ def check_job(job_id)
     rescue Faraday::Error::ClientError => e
       say_error_and_exit e.message
     end
+
+    private
+
+    def zat_details(response)
```
fvzzy: Maybe `zat_file_contents`

lib/zendesk_apps_tools/deploy.rb
keitheous: Agreed

spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -44,23 +37,48 @@ def get_value_from_stdin(text)
...
+        allow(subject).to receive_message_chain(:cache, :save)
+
         expect(subject.find_app_id).to eq(99)
       end
     end
   end
+
+  describe '#check_job' do
+    let(:failed_api_response_body) { { 'status' => 'failed', 'message' => 'You failed!' }.to_json }
+    let(:completed_api_response_body){ { 'status' => 'completed', 'app_id' => 2}.to_json }
```
fvzzy: Is `app_id` required here? Also perhaps make these two responses look the same except for the status code?

fvzzy:
Minor comments but otherwise 👍

spec/lib/zendesk_apps_tools/deploy_spec.rb
keitheous: App id is not required. Good catch.

About making the two responses look the same,
the failed message from response.body is captured and logged on the CLI, where as the one for completed isn't used anywhere in the code.

I'll have to go with this
```
let(:failed_api_response_body) { { 'status' => 'failed', 'message' => 'You failed!' }.to_json }

let(:completed_api_response_body){ { 'status' => 'completed'}.to_json } }
```

PR #270, opened 2018-12-19
[AF-1465] Lockdown gem versions to support Ruby Version 2.0.0 & 2.0.1

keitheous:
WIP PR to test approach

/cc @zendesk/vegemite :v:

### Tasks
- [ ] Test this manually

### Risks
* low - lockdown third level dependencies to support ruby 2.0.0 and 2.0.1


Gemfile
```diff
@@ -1,3 +1,11 @@
 source 'https://rubygems.org'
 
 gemspec
+
+gem 'rb-inotify', '~> 0.9', '>= 0.9.7'
+gem 'daemons', '~> 1.0', '>= 1.0.9'
+gem 'concurrent-ruby', '~> 1.1.3'
+# Bundler does not take current ruby version into account.
```
danielbreves: Maybe say `required_ruby_version` instead of current?

keitheous: Also as discussed IRL, no need to provide a range. Just directly reference the patch version in `~>`. replace first parameter with the second. no need 2 params

keitheous: Hey @danielbreves ,

I discovered something about `daemons` so I had to go with
```
gem 'rb-inotify', '~> 0.9.7'
gem 'concurrent-ruby', '~> 1.1.3'
gem 'daemons', '>= 1.0.9', '<= 1.3.0'
```

because current we rely on `demons` version `1.0.9` and `1.3.0`

danielbreves: 😈

hoangtrvu:
Closing in favour of https://zendesk.atlassian.net/browse/AF-1478

PR #273, opened 2019-01-03
[AF-1466] Ensure app id is fetched from a valid endpoint

fvzzy:
:v:

/cc @zendesk/vegemite

### Description
When `zat update` is run without an `app_id` stored in the `.zat` file, ZAT attempts to retrieve the ID from the deprecated `/api/apps.json` endpoint.

This change fixes `zat update` to instead search the (significantly more lightweight/efficient) installations endpoint, i.e. `/api/#{product_name}/apps/installations.json`. 

### Tasks
- [ ] Write tests

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1466

### Risks
* Low - ZAT update still doesn't work when an app name is manually input.


spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -6,46 +6,55 @@
...
       allow(subject).to receive(:get_value_from_stdin) { app_name }
-      allow(subject).to receive_message_chain(:cached_connection, :get, :body) { api_response_with_app_ids }
-
+      allow(subject).to receive_message_chain(:cached_connection, :get, :body) { api_response }
       subject
     end
 
+    context 'is unable to get a valid api response' do
+      it 'errors and exits system' do
+        subject = mocked_instance_methods_and_api('random_app_name', {})
```
keitheous: as discussed in person, this might be better as a let object to indicate that it is an invalid api response

spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -6,46 +6,55 @@
...
+        installations: [
+          { id: 1, app_id: 22, settings: { name: 'notification_app' }},
+          { id: 2, app_id: 99, settings: { name: 'time_tracking_app' }}
         ]
       }.to_json
-    end
+    }
 
-    define_method(:mocked_instance_methods_and_api) do |app_name|
+    define_method(:mocked_instance_methods_and_api) do |app_name, api_response|
```
keitheous: This might work with def method_name

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -62,26 +62,23 @@ def upload(path)
...
-        )
+      if installations_json.empty?
+        say_error_and_exit "Unable to retrieve installations. Please check your internet connection."
+      else
+        installation = json_or_die(installations_json)['installations'].find { |i| i.dig('settings', 'name') == app_name }
       end
 
-      cache.save 'app_id' => app['id']
-      app['id']
+      say_error_and_exit "App not found. Please verify that your credentials, subdomain, and app name are correct." unless installation
```
keitheous: might want to consider as a block as it extends to far to the right

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -62,26 +62,23 @@ def upload(path)
...
-
-      unless app
-        say_error_and_exit(
-          "App not found. " \
-          "Please verify that your credentials, subdomain, and app name are correct."
-        )
+      if installations_json.empty?
+        say_error_and_exit "Unable to retrieve installations. Please check your internet connection."
+      else
+        installation = json_or_die(installations_json)['installations'].find { |i| i.dig('settings', 'name') == app_name }
```
keitheous: Travis is failing here because ruby 2 2.1 and 2.6.6 doesn't know what dig is yet.

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -62,26 +62,27 @@ def upload(path)
       say_error_and_exit e.message
     end
 
-    def find_app_id
+    def find_app_id(product_name = 'v2') # use the v2 endpoint if no product name is provided
```
vaevictis: What are products in this context? Might be my lack of knowledge of ZAT, but this looks like a framework version to me. If that's the case, maybe call that variable `framework_version`?

spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -6,46 +6,56 @@
...
     context 'user inputs an app name that is in api response' do
       it 'returns app id 22 for notification_app' do
-        subject = mocked_instance_methods_and_api('notification_app')
+        subject = mocked_instance_methods_and_api('notification_app', api_response_with_installations)
         allow(subject).to receive_message_chain(:cache, :save)
 
         expect(subject.find_app_id).to eq(22)
       end
 
       it 'returns app id 99 for notification_app' do
```
vaevictis: There's some confusion here between the 2 `it` blocks and the descriptive string that follows. "returns app id 99 for **notification_app**" and `mocked_instance_methods_and_api('**time_tracking_app**')`. `#Fixme`?

lib/zendesk_apps_tools/deploy.rb
fvzzy: Products are `support`, `chat` etc. Technically those references are to a `v3` endpoint. We fallback to the `v2` endpoint if no product is specified as the response there helpfully looks identical to the v3 ones

spec/lib/zendesk_apps_tools/deploy_spec.rb
keitheous: this was my mistake sorry. I wrote the previous pr and i accidentally left this in it.

fvzzy: Good catch, thanks

spec/lib/zendesk_apps_tools/deploy_spec.rb
```diff
@@ -6,46 +6,56 @@
...
+        invalid_api_response = {}
+        subject = mocked_instance_methods_and_api('notification_app', invalid_api_response)
+
+        expect(subject).to receive(:say_error).with(
+          "Unable to retrieve installations. Please check your internet connection."
+        )
+        expect { subject.find_app_id }.to raise_error(SystemExit)
+      end
+    end
+
```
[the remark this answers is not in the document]
keitheous: Owh yes of course 👍

lib/zendesk_apps_tools/deploy.rb
```diff
@@ -62,26 +62,27 @@ def upload(path)
...
+      installations_json = cached_connection.get("/api/#{product_name}/apps/installations.json").body
 
-      app =
-        unless all_apps_json.empty?
-          json_or_die(all_apps_json)['apps'].find { |app| app['name'] == app_name }
-        end
+      if installations_json.empty?
+        say_error_and_exit "Unable to retrieve installations. Please check your internet connection."
+      else
+        installation = json_or_die(installations_json)['installations'].find {
```
token-cjg: If `json_or_die(installations_json)['installations']` is `nil`, this is throwing an exception, which probably should be handled.

Alternatively, one could have a check about this for

`installations = json_or_die(installations_json)['installations']`
`installation = installations.nil? ? nil : installations.find { <your code> }`

PR #277, opened 2019-03-06
Updated 'zat new' template

bryan-flynn-zd:
:v:

/cc @zendesk/vegemite

### Description
The template used to generate iframe.html when running `zat new` needs updating. The 'u-gamma' class attribute, for example, is no longer part of Zendesk Garden. Also, the 'bedrock' CSS library pulled in resets all styling and probably isn't the best demonstration of Garden.

Changed H2 text styling using Utilities CSS class settings. Changed relevant links as comments.

### Risks
* [low] Switching out existing Bedrock CSS library for Utilities library. Updated H2 class attributes.


app_template_iframe/assets/iframe.html
```diff
@@ -1,16 +1,19 @@
...
 <html>
 <head>
   <meta charset="utf-8">
-  <!-- http://garden.zendesk.com -->
-  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@zendeskgarden/css-bedrock@7.0" type="text/css">
+  <!--   See "Using Zendesk Garden":
+    https://developer.zendesk.com/apps/docs/developer-guide/setup#using-zendesk-garden
+    https://garden.zendesk.com/css-components/utilities/typography/
+   -->
+ <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@zendeskgarden/css-utilities@4/dist/index.min.css">
```
ocke: Any reason you removed bedrock from the stylesheet?

bryan-flynn-zd: It resets all default stylings, including browser ones I  believe, which I was thinking for a starting point for an app might be kind of jarring.

bryan-flynn-zd: @ocke +1 then and a merge? The default template using 'u-gamma' is incorrect now as there is no 'u-gamma' in Garden anymore. Bringing in 'Utilities' was to do something similar as before on the H2 block.

sarmadsangi: @bryan-flynn-zd I feel we should at least have css resets (normalize.css) in there. Bedrock helps inject [normalize.css](https://github.com/necolas/normalize.css/), default zendesk prefered fonts etc.

bryan-flynn-zd: Thanks @sarmadsangi for that additional info. I added back Bedrock. I also included both libraries separately with Bedrock first -- I didn't want Bedrock to be loaded after Utilities in one JSDelivr blob. I also made the links to be version specific, just to avoid any possible inconsistencies between versions.

app_template_iframe/assets/iframe.html
```diff
@@ -1,16 +1,21 @@
...
   <meta charset="utf-8">
-  <!-- http://garden.zendesk.com -->
-  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@zendeskgarden/css-bedrock@7.0" type="text/css">
+  <!--   See Using Zendesk Garden:
+    https://developer.zendesk.com/apps/docs/developer-guide/setup#using-zendesk-garden
+    https://garden.zendesk.com/css-components/bedrock/
+    https://garden.zendesk.com/css-components/utilities/typography/
+   -->
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@zendeskgarden/css-bedrock@7.0.20/dist/index.min.css">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@zendeskgarden/css-utilities@4.2.1/dist/index.min.css">
```
hoangtrvu: @pmgrove recently mentioned to me that we should use `bedrock` for `zat new`, not sure about `utilities` though. cc @sostopher and @ocke for thoughts too.

bryan-flynn-zd: I added Bedrock back in but kept Utilities, so `u-semibold u-fs-xl` could be used.

ocke: You can make this 1 request (creating the connection is the slowest part of downloading an app).

`<link rel="stylesheet" href="https://cdn.jsdelivr.net/combine/npm/@zendeskgarden/css-bedrock@7.0.20,npm/@zendeskgarden/css-utilities@4.2.1">`

ocke:
LGTM besides making it 1 request

PR #293, opened 2019-05-16
ZAT on windows, Approach 1, Part 3.  Add logic to curl new zam endpoint to validate svgs

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
Depends on https://github.com/zendesk/zendesk_apps_support/pull/225, which depends on https://github.com/zendesk/zendesk_app_market/pull/3521.

### Tasks
- [ ] Write tests

### References

### Risks
* [HIGH | medium | low] Does it work on windows?
* [HIGH | medium | low] Does it work in the different products (Support, Chat)?
* [HIGH | medium | low] Are there any performance implications?
* [HIGH | medium | low] Any security risks?
* [HIGH | medium | low] What features does this touch?

### DIFF
--- a/lib/zendesk_apps_tools/command.rb
+++ b/lib/zendesk_apps_tools/command.rb
@@ -88,6 +88,8 @@ module ZendeskAppsTools
       require 'execjs'
       check_for_update
       setup_path(options[:path])
+
+      # run usual validation checks
       begin
         errors = app_package.validate(marketplace: false)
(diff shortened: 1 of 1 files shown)


token-cjg:
Closing due to lack of activity + due to the fact that ZAT is now in maintenance mode.  Certainly however we can return to this if there is demand for this functionality in ZAT.

PR #295, opened 2019-06-12
Allow port to be set in the asset tags for theme previews.

TruRooms:
:v:

/cc @zendesk/vegemite

### Description
While setting a non standard port, scripts for the template were still getting injected with the standard port. Updated to respect the :port option when it is set using the standard options set and inject the appropriate script/style tags into the doc head.

### Tasks
- [ x] Include comments/inline docs where appropriate
- [ x] Write tests

### Risks
* [medium] Does it work on windows? Unable to verify
* [low] Does it work in the different products (Support, Chat)? Only for the form preview
* [low] Are there any performance implications? Optional port change should not effect performance.
* [medium] Any security risks? Allowing script tag to follow the port set from the CLI.
* [low] What features does this touch? The template live preview server.

### DIFF
--- a/lib/zendesk_apps_tools/theme.rb
+++ b/lib/zendesk_apps_tools/theme.rb
@@ -102,17 +102,17 @@ module ZendeskAppsTools
       def inject_external_tags(head_template)
         live_reload_script_tag = <<-html
           <script type="text/javascript">
-            RACK_LIVERELOAD_PORT = 4567;
+            RACK_LIVERELOAD_PORT = #{options[:port]};
           </script>
-          <script src="http://localhost:4567/__rack/livereload.js?host=localhost"></script>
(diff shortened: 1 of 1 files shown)


augustocravosilva:
Thanks for the PR!

Looks like there's still 2 hardcoded references to port '4567', would you be able to update that too?

https://github.com/zendesk/zendesk_apps_tools/blob/4604c686e5fd1816fe5a285493cd979f7a2f3d9e/lib/zendesk_apps_tools/theming/common.rb#L12
https://github.com/zendesk/zendesk_apps_tools/blob/4604c686e5fd1816fe5a285493cd979f7a2f3d9e/lib/zendesk_apps_tools/theme.rb#L145

Romuloux:
@augustocravosilva Good catch, references have been updated.

token-cjg (approved):

PR #305, opened 2019-07-18
[AF-1656] Add safe bump tasks and use these tasks when publishing zat

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
Adds new tasks `bundle exec rake bump:patch|minor|major:safe`.

These tasks bump the version of zat, without doing a bundle update.

### Tasks
- [ ] Write tests

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1656

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,5 @@
 require 'bundler/gem_tasks'
+require 'rake'
 require 'rake/clean'
 require 'cucumber/rake/task'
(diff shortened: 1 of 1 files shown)


token-cjg:
Rejecting this in favour of the other changes attached to https://zendesk.atlassian.net/browse/AF-1656 .

PR #306, opened 2019-07-24
[AF-1688] Automatically update version on merge (provided tests green)

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
Automatically updates the version on merge (provided tests green).

Similarly to https://github.com/zendesk/zendesk_app_framework/pull/852, if one merges a PR to zat with message major, minor, or other, this change will:

* `bump major|minor|patch --no-bundle` # to safely bump zat

### References
* JIRA: https://zendesk.atlassian.net/browse/AF-1688

### Risks
* (Medium) Might muddy up our versioning and release history for zat.


scripts/build_after_merge.sh
```diff
@@ -0,0 +1,36 @@
...
+if [[ $LAST_MSG =~ '[major]' ]]; then
+  BUMP='major'
+elif [[ $LAST_MSG =~ '[minor]' ]]; then
+  BUMP='minor'
+else
+  BUMP='patch'
+fi
+
+echo '*** Updating version'
+VERSION=$(bump $BUMP --no-bundle)
```
hoangtrvu: Just checking, does `bump` create a git tag as part of this process?

token-cjg: That is the intent, yes.  On checking this, I realised that it didn't work.  I've made a renewed attempt at getting this right in the latest commit.

token-cjg:
We are no longer using Travis, and ZAT is now in maintenance mode.

If we do want to eventually implement this, we will need to carefully decide how to do this with GH actions.

PR #326, opened 2019-12-02
Update REGEX for staging

ocke:
:v:

/cc @zendesk/vegemite

### Description
`zd-staging.com` was changed to `zendesk-staging.com` a long time ago. This old regex means `?zat=true` does not function correctly on staging (more specific the reload doesn't work).

### References

### Risks
* [none] only affects staging

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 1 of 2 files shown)


ocke:
Should we add something for sell here?

PR #327, opened 2019-12-03
Use ZAS v4.23.0 in ZAT BUMP v3.6.0

keitheous:
/cc @zendesk/vegemite @zendesk/dingo @jalee-zendesk 

### Description
Bumping ZAS to v2.26.0. In the meantime, bumping ZAT also in the process v3.6.0

### Risks
* [low] Only bumping ZAS for ZAM not for ZAT

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    zendesk_apps_tools (3.5.1)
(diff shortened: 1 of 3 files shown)


keitheous:
As discussed with Mehul in person, there will be no change in the developer experience when developers use the new ZAT bump.

ocke:
If this is only a ZAM change, why do we need to bump ZAT?

PR #328, opened 2019-12-03
[APPS-4945]Use ZAS v4.27.0 and BUMP ZAT v3.6.1

keitheous:
/cc @zendesk/dingo 

### Description
[JIRA story](https://zendesk.atlassian.net/browse/APPS-4945).
In this implementation, I have added a before_install script `gem install bundler -v 1.17.3` because that is the current default bundler version that supports ruby version 2.1 through to 2.4

Screen Shot 2020-01-10 at 10 47 03 am

Also, we need ZAS bumped to support custom objects when it becomes available.

### Risks
* [low] Bumping ZAS in ZAT.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 ---
 language: ruby
 dist: trusty
+before_install:
+  gem install bundler -v 1.17.3
 rvm:
   - 2.1
   - 2.2.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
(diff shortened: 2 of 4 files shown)


keitheous:
@Mehul-Porwal 
> BTW does this change the developer experience? Any changes to ZAT/app submission that should be documented on developer.zendesk.com?
> Looks good, but pending confirmation on no new documentation is required before merging.

As discussed in person. This does not change the developer experience. This work is mainly for custom objects.

@ocke 
> If this is only a ZAM change, why do we need to bump ZAT?

As for now this change only affects ZAM, but when custom objects is available, developers want to developers to be able to do `zat validate`

PR #334, opened 2020-01-29
Bump ZAS in ZAT

Mehul-Porwal:
:v:

/cc @zendesk/vegemite

### Description
Look at title.

### Risks
* Low. Local development may break with this version(4.28.0) installed.


token-cjg:
Are we bumping ZAS in ZAT still? Or just dropping support for ruby 2.1 and 2.2?

PR #338, opened 2020-05-14
Bump of zendesk_apps_support dependency

mikhail-ev:
:v:

/cc @zendesk/vegemite

### Description
Bumping zendesk_apps_support version in order to get new available location `visit_editor` from zendesk_apps_support's latest release

### Tasks
- [ ] Write tests

### References
* JIRA: https://zendesk.atlassian.net/browse/SFP-841

### Risks
* [low] Does it work on windows? Should not be affected
* [low] Does it work in the different products (Support, Chat)? Should not be affected
* [low] Are there any performance implications? No
* [low] Any security risks? No
* [low] What features does this touch? Locations update

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -12,7 +12,7 @@ PATH
       sinatra-cross_origin (~> 0.3.1)
       thin (~> 1.7.2)
       thor (~> 0.19.4)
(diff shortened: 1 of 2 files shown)


sarmadsangi:
@mikhail-ev looks like you need to update gem lock file too.

PR #378, opened 2022-04-11
[VEG-1234] Add ability to mount ZAT in a docker container for testing

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
We'd like to be able to run `zat` in a docker container, so as to be able to work around environment constraints, as well as test various things that can't be tested just by running zat natively.

### Tasks
- [x] Include comments/inline docs where appropriate

### References
* JIRA: https://zendesk.atlassian.net/browse/VEG-1234

### Risks
* Low. Adds a new way of running ZAT, does not touch upon anything that is included in the gemspec.


scripts/invoke.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# When running the zat container, mount the current directory to /app
+# so that zat has access to it.
+docker run -it --rm -v $(pwd):/usr/src/app zat $@
```
cl8au, line 5: hmm... I thought you already copy the src folder as part of the docker build. Why do you need to mount it again?

In addition, if you need to mount it for running unit tests changes, do you need to follow the Zendesk environment variable to use that to point the zat repository under your code path?

zidmus (approved):
LGTM

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ruby:2.6
```
cl8au, line 1: should we use zendesk ruby base image not the public one?

scripts/invoke.sh
Frederick888: I think this is intended for use in app directories, hence it's mounting app source codes rather than ZAT code.

Though by the way I believe `$(pwd):/usr/src/app` and `$@` both need to be quoted:
- https://github.com/koalaman/shellcheck/wiki/SC2046
- https://github.com/koalaman/shellcheck/wiki/SC2068

scripts/compile.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
Frederick888, line 1: `#!/usr/bin/env bash` is a better shebang :)

scripts/invoke.sh
cl8au: oh, so $(pwd) indicating the application code base folder instead of zat project. I see now

zat.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [[ "$(docker images -q zat:latest 2> /dev/null)" == "" ]]; then
```
Frederick888, line 3: nit: can use `[[ -z "$(...)" ]]` instead

scripts/invoke.sh
token-cjg: ah, I see the issue. I need to alias this command as eg `zatDev`, or write a shortcut to mount it within ~/.zshrc or somesuch

basically the usecase is

```
cd mycoolapp
zatDev apps server
```

so we'd run the container and mount the mycoolapp directory within it, so that `zatDev apps server` notices when things change and updates the served app accordingly

This ask actually came from `zat themes preview`, we want to mount a theme and live poll for updates to the directory on the machine which we mount as a volume relative to the running container.

zat.sh
```diff
@@ -0,0 +1,8 @@
...
+  echo "ZAT Image not found, building..."
+  ./scripts/compile.sh
+fi
+
+./scripts/invoke.sh $@
```
Frederick888, line 8: nit: Consistently having EOL at the end of files would be an eye candy :)

zat.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [[ -z "$(docker images -q zat:latest 2> /dev/null)" ]]; then
+  echo "ZAT Image not found, building..."
+  ./scripts/compile.sh
+fi
+
+./scripts/invoke.sh $@
```
Frederick888, line 8: 1. `$@` here also needs to be quoted
2. Does this actually work? If I run this in an app directory, wouldn't it complain something like `./scripts/invoke.sh does not exist`?

token-cjg:
I've given this another once over, this should work now.

The main change is to build the gem and install it in the image, and then set the entrypoint to `zat`.  This allows the image to be run from anywhere.

token-cjg:
Yes, I think that makes sense.  So I should `EXPOSE 4254 4567` in the Dockerfile.  Thanks for the eagle eye!

cl8au:
> Yes, I think that makes sense. So I should `EXPOSE 4567` in the Dockerfile. (I believe that port `4254` is an example of a custom port which doesn't need to be supported.) Thanks for the eagle eye!



> Yes, I think that makes sense. So I should `EXPOSE 4567` in the Dockerfile. (I believe that port `4254` is an example of a custom port which doesn't need to be supported.) Thanks for the eagle eye!

Maybe this is a very nit question and again also feel free to ignore. If zat server supports custom port. should we also support it from your shell script to pass the argument as port if needed in order to let the docker container to expose the port?

token-cjg:
Yes, indeed we should, otherwise this will not work.  Thanks for the astute observation 🙇

scripts/invoke.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+PROGNAME="$(basename $0)"
+
+# Helper functions for guards
+error(){
+  error_code="$1"
+  echo "ERROR: $2" >&2
+  echo "($PROGNAME wrapper version: $VERSION, error code: $error_code )" &>2
```
Frederick888, line 9: Hmmm... where is this `$VERSION` from?

scripts/invoke.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit "$1"
+}
+check_cmd_in_path(){
+  cmd="$1"
+  which $cmd > /dev/null 2>&1 || error 1 "$cmd not found!"
```
Frederick888, line 14: nit: Prefer `command -v`. Ref https://stackoverflow.com/a/677212/2487227

token-cjg:
ha! yes that's an artifact of lazily copying code.  I will fix that.

Frederick888 (approved):

PR #382, opened 2022-04-20
Update Gemfile.lock

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
The Dependabot PRs did not bump Gemfile.lock, we should `bundle install` and then bump Gemfile.lock accordingly.

### References
* JIRA: tbd

### Risks
* Medium. Gemfile.lock changes lead to a degradation in ZAT functionality.


Frederick888 (approved):
Should be ok as long as CI doesn't complain :)

PR #389, opened 2022-05-18
Add bundler-audit and bundler-audit-fix

token-cjg:
### Description

Adds tools for auditing vulnerabilities for ZAT. This is important in
order to make sure that we ship secure code.

As part of this change I needed to bump the major version of Thor, which
is a gem used to handle the command line functionality for ZAT.

When running bundler-audit-fix one can run

`bundle exec bundler-audit-fix update .` `[1]`

This should address vulnerabilities identified by `bundler-audit`.

`[1]`: https://github.com/nobuyo/bundler-audit-fix

:v:

/cc @zendesk/vegemite

### Tasks
- [ ] Write tests

### References
* tbd

### Risks

Medium. Might break the command line functionality of ZAT as Thor, the library which powers the CLI, has been bumped by a major version.  This should be easy to check though -- if one builds the gem manually and run `./bin/zat version`, then things should still work.

### DIFF
--- a/zendesk_apps_tools.gemspec
+++ b/zendesk_apps_tools.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.3'
   s.required_rubygems_version = '>= 1.3.6'
 
-  s.add_runtime_dependency 'thor',        '~> 0.19.4'
+  s.add_runtime_dependency 'thor',        '~> 1.2.1'
   s.add_runtime_dependency 'rubyzip',     '>= 1.2.1', '< 2.4.0'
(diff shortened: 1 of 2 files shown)


zendesk_apps_tools.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.3'
   s.required_rubygems_version = '>= 1.3.6'
 
-  s.add_runtime_dependency 'thor',        '~> 0.19.4'
+  s.add_runtime_dependency 'thor',        '~> 1.2.1'
```
cl8au, line 18: this is a major version change which I guess require us to test it before release it out :)

cl8au (approved):

PR #396, opened 2022-08-02
Upgrade rack to 2.2.4

txu-gsd:
:v:

/cc @zendesk/vegemite

### Description
We need to upgrade Rack to 2.2.4 to prevent Arbitrary Code Injection exist in the current version

### Tasks
- [ ] Write tests

### References
* JIRA: N/A

### Risks

* [Low] This is only a patch update, no breaking changes


lib/zendesk_apps_tools/version.rb
```diff
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module ZendeskAppsTools
-  VERSION = '3.8.13'
+  VERSION = '3.8.14'
```
Frederick888, line 3: Need to update Gemfile.lock as well

Frederick888 (approved):

zidmus (approved):

PR #401, opened 2022-08-18
Bump Ruby version and ZAT version

Frederick888:
### Description

- Bump required Ruby to >= 2.6
- Bump required RubyGems version to 3.0.0
- Bump version to 3.9.0

### Tasks
- [ ] Write tests

### References
* JIRA: N/A

### Risks

Low: We've actually been using nokogiri 1.13.x for a while so ZAT
already requires Ruby >= 2.6.

### DIFF
--- a/zendesk_apps_tools.gemspec
+++ b/zendesk_apps_tools.gemspec
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
   s.summary     = 'Tools to help you develop Zendesk Apps.'
   s.description = s.summary
 
-  s.required_ruby_version = '>= 2.3'
+  s.required_ruby_version = '>= 2.6'
   s.required_rubygems_version = '>= 1.3.6'
 
   s.add_runtime_dependency 'thor',        '~> 0.19.4'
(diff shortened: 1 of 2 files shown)


token-cjg (approved):

zendesk_apps_tools.gemspec
```diff
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
   s.summary     = 'Tools to help you develop Zendesk Apps.'
   s.description = s.summary
 
-  s.required_ruby_version = '>= 2.3'
+  s.required_ruby_version = '>= 2.6'
   s.required_rubygems_version = '>= 1.3.6'
```
token-cjg, line 16: Should we update the required_rubygems_version too?

Frederick888: We can, but I'm not sure which version I should bump it to.

Locally I'm using Ruby 2.6.9, and it comes with `gem 3.0.3.1`. So something like `>= 3.0.0`?

cl8au (approved):

zendesk_apps_tools.gemspec
Frederick888: Pushed 👌

Issue #364, opened 2021-07-26
Not reloading on Guide's theme preview


suzu2469:
The preview of my theme is not automatically reflected during theme development in Guide.
Updating my theme files does not automatically upload them and the `zat theme preview` console log is unresponsive.
I started experiencing this symptom after updating to macOS BigSur.
What should I do?

````
OS: macOS BigSur 11.4

❯ ruby -v
ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-darwin20].
```

RamChilla:
I'm having the same problem. The theme preview works on the first load. I go back to my editor, make a change, and go back to the browser and refresh, but no changes reflected.

I have to start and stop the preview after every change I would like to test.

alexk-lb:
Having the same problem, looks like it's been this way for several months. For now the only workaround I've found is the one @RamChilla stated above: starting and stopping `zat theme preview` every time I make a change.

alexk-lb:
I spent a bit more time looking at this and in my case, the live-reload was failing on tokenized assets (e.g. `$assets-customfont-woff`) in my stylesheet. Commenting out those assets and/or any place they were used as a value in a CSS property fixed the live-reload feature.

glook:
I made some changes that can help with this issue https://github.com/zendesk/zendesk_apps_tools/pull/374

token-cjg:
Hi all, thanks for the issue report for zendesk_apps_tools.

@glook thank you for submitting a candidate fix for this.

https://github.com/zendesk/zendesk_apps_tools/pull/374 has now been merged [and released with tag 3.8.7](https://rubygems.org/gems/zendesk_apps_tools/versions/3.8.7), please advise if this is still an issue while working with `zat theme preview`.

PaddyMurphy:
@token-cjg I was having the same issue `zat theme preview` works until I make a change (I’m using zendesk_apps_tools 3.8.8).

I can confirm the @alexk-lb solution worked for me. I removed all references to $assets in style.css and it works perfectly.

running macos 12.3.1 and ruby 2.6.8p205

evenfrost:
I'm having an issue when using `$assets-` helpers with fonts completely hangs the local server, and the Zendesk HC page never finishes loading.  In the Network tab I see a bunch of pending requests from `localhost:4567` that just stay there until I forcefully kill the server and start it again in the new tab.

```
@font-face {
	font-family: 'Roboto';
[...]

PR #404, opened 2022-10-30
Bump listen to 3.7.1

token-cjg:
:v:

/cc @zendesk/vegemite

### Description
[It has been reported](https://github.com/zendesk/zendesk_apps_tools/issues/364#issuecomment-1296287312) that bumping listen to 3.7.1 should fix this issue: https://github.com/zendesk/zendesk_apps_tools/issues/364.

Compare diff: https://github.com/guard/listen/compare/v2.10.1...v3.7.1

### Tasks
- [ ] Write tests

### References
* JIRA: https://zendesk.atlassian.net/browse/tbd
* https://github.com/zendesk/zendesk_apps_tools/issues/364

### Risks
High. Might break `zat server` and `zat theme preview` functionality; anything that live reloads from the file system in order to serve an object to a target account.  However, it should be easy enough to check that the functionality still works after this gem bump.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,7 @@ PATH
       execjs (~> 2.7.0)
       faraday (~> 0.17.5)
       faye-websocket (>= 0.10.7, < 0.12.0)
-      listen (~> 2.10)
+      listen (~> 3.7.1)
       rack-livereload
       rubyzip (>= 1.2.1, < 2.4.0)
       sinatra (>= 1.4.6, < 2.3.0)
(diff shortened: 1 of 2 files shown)


Frederick888 (approved):
LGTM.

Probably has something to do with https://github.com/guard/listen/pull/479? Anyway the `Listen.to()` API seems compatible and 3.7.1 requires Ruby >= 2.4.0, which is well below our requirement as well.

Still would appreciate it if you can test it out before merging 👍

PR #433, opened 2026-01-20
APPS-7934: Upgrade `rack` to `3.2`

zendesk-poonam:
:v:

/cc @zendesk/vegemite

### Description
Upgrade `rack` to `3.2`

### Tasks
- [ ] Write tests

### References
* JIRA: https://zendesk.atlassian.net/browse/APPS-7934

### Risks

* [ low] `rack` version upgrade

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -32,9 +32,9 @@ jobs:
       - name: before_script
         run: |
           gem update --system 3.3.22
-          bundle config set frozen true
           bundle config set --local path 'vendor/cache'
           bundle install --jobs=3 --retry=3
+          bundle config set frozen true
           gem build zendesk_apps_tools.gemspec
           GEM_VERSION="$(sed -En "s/^.*VERSION = '([[:digit:]]+\\.[[:digit:]]+\\.[[:digit:]]+(\\.(beta|pre|rc)([[:digit:]]+|\\.?[[:digit:]]+)?)?)'$/\\1/p" lib/zendesk_apps_tools/version.rb)"
           gem install "./zendesk_apps_tools-$GEM_VERSION.gem"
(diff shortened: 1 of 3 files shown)


zendesk_apps_tools.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'rubyzip',     '>= 1.2.1', '< 2.4.0'
   s.add_runtime_dependency 'thin',        '~> 2.0'
   s.add_runtime_dependency 'sinatra',     '~> 4.1'
-  s.add_runtime_dependency 'rack',        '~> 3.0'
+  s.add_runtime_dependency 'rack',        '~> 3.2'
```
zendesk-poonam, line 22: Upgraded to version `3.2`, as it is compatible with all three Ruby versions: `3.1, 3.2, and 3.3`

.github/workflows/actions.yml
```diff
@@ -32,9 +32,9 @@ jobs:
       - name: before_script
         run: |
           gem update --system 3.3.22
-          bundle config set frozen true
           bundle config set --local path 'vendor/cache'
           bundle install --jobs=3 --retry=3
+          bundle config set frozen true
```
zendesk-poonam, line 37: Reordered so bundler can actually download the new rack `3.2.4` gem instead of being stuck with only what's in the old cache.
Earlier it was failing [here.](https://github.com/zendesk/zendesk_apps_tools/actions/runs/21168344239/job/60878581912?pr=433)

aishkan: `bundle config set frozen true` should be set before bundle install right so the same lockfile is used ?

aishkan: Can we try removing `bundle config set --local path 'vendor/cache'` and setting bundler-cache:true [here](https://github.com/zendesk/zendesk_apps_tools/pull/433/files#diff-2702f6f7d10a1bffbc8596d1a9925b1234bc4c7a6ff4c21ca6b34c13aa75780eR21) instead ?

zendesk-poonam:
We've disabled dependabot alerts for non-runtime dependencies for ZAT, so this is not needed anymore.

